PR #2, opened 2020-07-20
Add initial existing conventions to docs/

thejcannon:


docs/Coding-Conventions.md
```diff
@@ -0,0 +1,906 @@
+NOTE: This guide is a Work In Progress!
+
+[[_TOC_]]
```
thejcannon, line 3: This doesn't work in GitHub-world. 😢

thejcannon:
I stripped anything that referenced internal links/teams/etc...

docs/Coding-Conventions.md
mshafer-NI: it does on GitHub pages - which uses Kramdown

thejcannon:
FYI @zhindes

docs/Coding-Conventions.md
thejcannon: Right now, my plan is to download a VS Code Extension to do this work for me.

Eventually I think having a GitHub Action validate the TOC on PRs would be 🔥

thejcannon: We can also use a GitHub Action to just _generate_ the TOC too (at the cost of additional commits, which... 🤷‍♂️ )

thejcannon:
I think I'd like one more approval here to make sure I didn't miss anything obvious. @neilvana @jryckman?

PR #3, opened 2020-07-23
Add CODEOWNERS validation to PR

thejcannon:


mshafer-NI:
Could you touch the CODEOWNERS file so that the validation is triggered and we can have it in this PR that it passes?

thejcannon:
Does GitHub trigger the pre-creation actions on a PR?

thejcannon:
I kinda hope not, as I could `echo` a secret 😄

ScotSalmon:
If it's like Azure Pipelines it might just filter out the secret from the echo'd output.

.github/workflows/Validate-CODEOWNERS.yml
```diff
@@ -0,0 +1,20 @@
...
+    paths:
+      - '.github/CODEOWNERS'
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: mszostok/codeowners-validator@v0.4.0
```
alejandro5042, line 16: Nice to have: Maybe include an easy link to the check? https://github.com/mszostok/codeowners-validator

mshafer-NI:
> Does GitHub trigger the pre-creation actions on a PR?

I'm not sure where pre-creation is coming from...

You've set `on: pullrequest` - https://docs.github.com/en/actions/configuring-and-managing-workflows/configuring-a-workflow#triggering-a-workflow-with-events

thejcannon:
@ScotSalmon 

Perhaps they weren't "secrets" in the pipeline sense, but I've definitely used a pipeline which I'm changing in a PR to have the PR pipeline print out the contents of files with sensitive information 👀 

---

@mshafer-NI 

> I'm not sure where pre-creation is coming from...
> 
> You've set on: pullrequest, so I would expect this to run on PR's that touch the file you have it filtered to

pre-creation is the before-it-exists. Basically, would GitHub trigger an action for a PR in which the action is being created for that PR.

That being said, let's just try it 😄

thejcannon:
@mshafer-NI Doesn't look like it's the case. Reverting my change to CODEOWNERS

thejcannon:
FYI tried this out on my fork, and seemed to work 👍

PR #5, opened 2020-07-27
Add GitHub Pages Support

thejcannon:
This (along with a repo settings change) will have us serving our conventions using GitHub Pages.

You can see what the page will look like by going to my fork's sourcing of this change: https://thejcannon.github.io/ni-python-styleguide/


docs/index.md
```diff
@@ -0,0 +1,45 @@
+<!-- This is to make the GitHub Page for the repo look nice  -->
```
thejcannon, line 1: This file uses `Jekyll`: https://jekyllrb.com/ (as that is what GitHub Pages uses), and the Markdown itself is rendered via `kramdown`: https://kramdown.gettalong.org/index.html

Most of the code here is just to add a few niceties to make the document a bit more... professional. Additionally, I wanted to make the GitHub `/blob/` link to the conventions markdown not look... unprofessional 😄 (Just in case someone looks at that/sends a link/etc...)

Since it's a templating language (which aren't the easiest to read), I'll outline the important bits:
- Moved ToC support here by using a comment in the actual Markdown. The ToC is handled by `kramdown`'s ToC markdown extension
- Leveraged `kramdown`'s markdown extension for choosing a header's ID for convention headers (so they don't use the whole friggin' convention in the link 😓)
- Added link icons before each header which links to itself so people can easily copy the convention's direct link

docs/index.md
```diff
@@ -0,0 +1,48 @@
...
+        {%- else -%}
+            {%- assign code_block = true -%}
+        {%- endif -%}
+    {%- endif -%}
+
+    {%- assign first_token_char = first_token | slice: 0 -%}
+    {%- if auto_id and first_token_char == "#" and code_block == false -%}
+        {% comment %}Omit hyphens, as we need the whitespace to denote a header{% endcomment %}
+        {% assign header_text = line | split: first_token | last %}
+        {% assign header_id = line | split: "]" | first | split: "[" | last | replace: ".","_" | downcase %}
```
thejcannon, line 37: So the IDs look like this:
- https://thejcannon.github.io/ni-python-styleguide/#f
- https://thejcannon.github.io/ni-python-styleguide/#f_1
- https://thejcannon.github.io/ni-python-styleguide/#f_1_1

Some thoughts:
- Does lowercase letter or uppercase letter look better?
- Should we use underscore or dash?
- For the section/subsection should we also add the word/phrase? (E.g. `.../#f_1-indents`)

mshafer-NI: dashes vs underscores - dashes are harder to confuse with spaces when underline is involved, so I would suggest dashes.

include section word/phrase? - if I post a link to a subheader in a review, the phrase in the url can help the reader know that I'm poingint them to a relevant rule

mshafer-NI: upper vs lower - I would argue that it should match what is in place on the doc (which looks to be upper)

thejcannon: I agree, I'm going to change, but leave this open in case others want to discuss further.

ScotSalmon: I like the idea of having the link be more than just a section number if it's not too much trouble.

thejcannon: OK links are now:
- .../#F-formatting
- .../#F-1-indents
- .../#F-2-1

docs/Coding-Conventions.md
```diff
@@ -1,6 +1,6 @@
 NOTE: This guide is a Work In Progress!
 
-[[_TOC_]]
+<!-- TOC -->
 
 # Purpose of coding conventions
```
mshafer-NI, line 5: Suggestion, start Auto-ID/linking here - as purpose and how to read are likely things that we will have to point people at as well.

mshafer-NI: oh they are in the toc - just add the link button?

thejcannon: It's unfortunately not that simple. The current logic assumes the brackets in the header, as it makes the code simpler. Without that assumption, the Liquid templating magic that would need to take place can get hairy...

Those still have IDs, just not the little link icon.

I'm trying to play around with a library to do the link icon automatically, which would solve this issue.

thejcannon: OK I got AnchorJS working: https://github.com/bryanbraun/anchorjs

thejcannon:
I've made a request to see what we can/should use as the site's favicon. Stay tuned!

PR #8, opened 2020-07-27
Vendoring flakehell

thejcannon:
While we're waiting for flakehell to accept the PRs, we can vendor the dependency ourselves. Although this will increase our package size, and add a smidgen of complexity to our repo, it keeps our client's installation workflow simple.

Vendoring isn't new to the Python world. One example is build-systems having to vendor _all_ of their direct/indirect dependencies.

The changes of note can be found in the second commit onward. (First commit is the copying of flakehell files)

NOTE: This is not intended as a long-term solution.


mshafer-NI:
Testing done?

thejcannon:
`poetry install` correctly installs everything and running `ni-python-styleguide lint` on a dir does what I expect

PR #9, opened 2020-07-30
Rename links/text for repo rename

thejcannon:


README.md
```diff
@@ -1,4 +1,4 @@
-# ni-python-styleguide
+# NI Python Styleguide
```
alejandro5042, line 1: Minor: I think styleguide is actually two words: "NI Python Style Guide". I think it's OK if the repo name is "blah-styleguide" however.

PR #11, opened 2020-08-07
Add PR action and associated fixes to get it working

thejcannon:
This will allow us to runs several checks on the code and files when a new PR comes in. It also leverages the tool to lint the tool which I think is admirable.

Non-yaml changes are those required to get everything up and running. They are:

- PR \#92 to flakehell was made and applied to the vendored copy to allow subdirectory exclusions
- Since we're vendoring flakehell, we also need to define its `entry_points` (`plugins` in `poetry`)
- The lock file was out of date, so I updated it
- The build system was no longer valid so I updated it with what the docs recommend

### DIFF
--- a/.github/workflows/PR.yml
+++ b/.github/workflows/PR.yml
@@ -1,0 +1,46 @@
+name: PR
+
+on:
+  pull_request:
+    branches:
+      - master
+    paths:
+      - ni_python_styleguide
+      - poetry.lock
+      - pyproject.toml
+
+jobs:
+  pr:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+    - name: Check Poetry Metadata
+      uses: abatilo/actions-poetry@v1.5.0
+      with:
+        python_version: 3.5.0
+        poetry_version: 1.0
+        args: check
+
+    # @TODO: This is a workaround for there not being a way to check the lock file
+    #   See: https://github.com/python-poetry/poetry/issues/453
+    #   NOTE: We're taking advantage of the fact that "abatilo/actions-poetry"
+    #     has been run, installs poetry globally, and uses pyenv
+    - name: Check for lock changes
+      run: |
+        pyenv latest local 3.5.0
+        python -c "from poetry.factory import Factory; l = Factory().create_poetry('.').locker; exit(0) if l.is_locked() and l.is_fresh() else exit(1)" && echo 'OK'
+
+    - name: Install the Package
+      uses: abatilo/actions-poetry@v1.5.0
+      with:
+        python_version: 3.5.0
+        poetry_version: 1.0
+        args: install
+
+    - name: Lint the Code
+      uses: abatilo/actions-poetry@v1.5.0
+      with:
+        python_version: 3.5.0
+        poetry_version: 1.0
+        args: run ni-python-styleguide lint 
--- a/.github/workflows/PR.yml
+++ b/.github/workflows/PR.yml
@@ -24,11 +24,9 @@ jobs:
 
       # @TODO: This is a workaround for there not being a way to check the lock file
       #   See: https://github.com/python-poetry/poetry/issues/453
-      #   NOTE: We're taking advantage of the fact that "abatilo/actions-poetry"
-      #     has been run, installs poetry globally, and uses pyenv
       - name: Check for lock changes
         run: |
-          pyenv latest local 3.5.8
+          pip install poetry==1.0
           python -c " \
             from poetry.factory import Factory; \
             locker = Factory().create_poetry('.').locker; \
--- a/.github/workflows/PR.yml
+++ b/.github/workflows/PR.yml
@@ -24,8 +24,11 @@ jobs:
 
       # @TODO: This is a workaround for there not being a way to check the lock file
       #   See: https://github.com/python-poetry/poetry/issues/453
+      #   NOTE: This isn't ideal as it using the system Python and dirtying the system
+      #     Python libraries.
       - name: Check for lock changes
         run: |
+          pip install poetry==1.0
           python -c "from poetry.factory import Factory; \
             locker = Factory().create_poetry('.').locker; \
             exit(0) if locker.is_locked() and locker.is_fresh() else exit(1)" \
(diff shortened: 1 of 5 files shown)


thejcannon:
(Internal link) I made a Work Item https://dev.azure.com/ni/DevCentral/_workitems/edit/1105640 to research linting our Convention docs themselves (at least the Python docs) which serves multiple purposes:

- Acts as a form of test of the tool
- Ensures our lint rules are behaving as expected
- Ensures our conventions are honest

It is, however, outside the scope of this PR. 😄

mshafer-NI:
We got a PR check!

```
/usr/bin/docker run --name cbbef60fcd563541428a49e0f47089e6a2_1dad0e --label 8118cb --workdir /github/workspace --rm -e INPUT_PYTHON_VERSION -e INPUT_POETRY_VERSION -e INPUT_ARGS -e INPUT_WORKING_DIRECTORY -e HOME -e GITHUB_JOB -e GITHUB_REF -e GITHUB_SHA -e GITHUB_REPOSITORY -e GITHUB_REPOSITORY_OWNER -e GITHUB_RUN_ID -e GITHUB_RUN_NUMBER -e GITHUB_ACTOR -e GITHUB_WORKFLOW -e GITHUB_HEAD_REF -e GITHUB_BASE_REF -e GITHUB_EVENT_NAME -e GITHUB_SERVER_URL -e GITHUB_API_URL -e GITHUB_GRAPHQL_URL -e GITHUB_WORKSPACE -e GITHUB_ACTION -e GITHUB_EVENT_PATH -e RUNNER_OS -e RUNNER_TOOL_CACHE -e RUNNER_TEMP -e RUNNER_WORKSPACE -e ACTIONS_RUNTIME_URL -e ACTIONS_RUNTIME_TOKEN -e ACTIONS_CACHE_URL -e GITHUB_ACTIONS=true -e CI=true -v "/var/run/docker.sock":"/var/run/docker.sock" -v "/home/runner/work/_temp/_github_home":"/github/home" -v "/home/runner/work/_temp/_github_workflow":"/github/workflow" -v "/home/runner/work/python-styleguide/python-styleguide":"/github/workspace" 8118cb:bef60fcd563541428a49e0f47089e6a2 check
Latest version is '3.5.0'
Downloading Python-3.5.0.tar.xz...
-> https://www.python.org/ftp/python/3.5.0/Python-3.5.0.tar.xz
Installing Python-3.5.0...
ERROR: The Python ssl extension was not compiled. Missing the OpenSSL lib?
```

.github/workflows/PR.yml
```diff
@@ -0,0 +1,46 @@
...
+          args: check
+
+      # @TODO: This is a workaround for there not being a way to check the lock file
+      #   See: https://github.com/python-poetry/poetry/issues/453
+      #   NOTE: We're taking advantage of the fact that "abatilo/actions-poetry"
+      #     has been run, installs poetry globally, and uses pyenv
+      - name: Check for lock changes
+        run: |
+          pyenv latest local 3.5.0
+          python -c "from poetry.factory import Factory; l = Factory().create_poetry('.').locker; exit(0) if l.is_locked() and l.is_fresh() else exit(1)" && echo 'OK'
```
alejandro5042, line 32: I'd recommend against `l` as a variable name. Instead, we should name it `poetry_lock_file` or something similar.

.github/workflows/PR.yml
```diff
@@ -0,0 +1,46 @@
...
+        with:
+          python_version: 3.5.0
+          poetry_version: 1.0
+          args: check
+
+      # @TODO: This is a workaround for there not being a way to check the lock file
+      #   See: https://github.com/python-poetry/poetry/issues/453
+      #   NOTE: We're taking advantage of the fact that "abatilo/actions-poetry"
+      #     has been run, installs poetry globally, and uses pyenv
+      - name: Check for lock changes
```
alejandro5042, line 29: Educate me :) What does this lock check do?

thejcannon:
I filed an issue about the `poetry` action failing to install Python: https://github.com/abatilo/actions-poetry/issues/32

.github/workflows/PR.yml
thejcannon: I'm torn. I agree, but also this is an exceptional case: We're passing Python code in via a string in a throwaway line. Other things we shouldn't do: Use semicolons to separate lines 😛, call `exit`...

Maybe compromise and call it `lock`?

.github/workflows/PR.yml
thejcannon: Makes sure the lock is correct w.r.t to `pyproject.toml` -> (You didn't change the pyproject dependencies manually and forget to `poetry lock`).

Generally speaking you shouldn't be editing the `pyproject.toml` directly, you should use the commands (`poetry add`, `poetry remove`, `poetry update`). But from a PR point-of-view, there's no way to know if they did things correctly or not, so checking is ideal.

.github/workflows/PR.yml
thejcannon: I ended up calling it `locker` since that's the name of the propterty.

thejcannon:
`flakehell` claims to support Python 3.5+, however it's not apparent to me whether or not they test using Python 3.5.

I'm going to try to bump the Python version to 3.6. This tool can be run in isolation if a package needs to support Python 3.5 (or in a different version of Python, as presumably we'll be running tests with different versions, so multiple version are available to us). It isn't ideal, but is a decent workaround. Alternatively we can try and make `flakehell` truly Python 3.5 compatible, but I think that's overboard.

thejcannon:
Look at that, the PR failed because the lock file wasn't up-to-date 🎉

.github/workflows/PR.yml
```diff
@@ -0,0 +1,41 @@
...
+jobs:
+  pr:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: actions/setup-python@v1
+        with:
+          python-version: 3.6
+      - uses: Gr1N/setup-poetry@v2
```
alejandro5042, line 21: Minor: Seems like a new line before this line is good.

PR #12, opened 2020-08-11
Test conventions

thejcannon:
Adding pytest tests which do some nice checks on the conventions themselves.

This also serves as a good first step towards testing that our conventions are honest in their codeblocks/demarcation of enforcement.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,0 +1,37 @@
+import pytest
+
+from ni_python_styleguide.__main__ import main as styleguide_main
+
+
+def pytest_collection_modifyitems(items):
+    for item in items:
+        item.add_marker(pytest.mark.filterwarnings("ignore::DeprecationWarning"))
+
+
+@pytest.fixture
+def styleguide(request):
+    """A fixture which when executed, runs the styleguide.
+
+    Args passed to the function are equivalent to CLI args,
+    and are automatically stringified.
+
+    The return value is whether the command succeeded or failed.
+
+    E.g. `styleguide("lint", ".")`
+    """
+
+    def runner(*args):
+        if args[0] == "lint":
+            # Use default formatting instead of the colored formatting
+            args = [
+                *args,
+                "--format",
+                "default",
+            ]
+
+        with pytest.raises(SystemExit) as exc_info:
+            styleguide_main(["ni_python_styleguide", *map(str, args)])
+
+        return exc_info.value.code == 0
+
+    return runner
--- a/tests/test_convention_doc/test_rules.py
+++ b/tests/test_convention_doc/test_rules.py
@@ -2,15 +2,31 @@
 
 import re
 
+import pytest
+
+from tests.test_convention_doc import doctypes
+
+
+@pytest.fixture(
+    scope="module",
+    params=[
+        pytest.param((index, rule), id=rule.identifier)
+        for subsection in doctypes.SUBSECTIONS
+        for index, rule in enumerate(subsection.rules)
+    ],
+)
+def enumerated_rule(request):
+    return request.param
+
 
 def test_rule_identifier_valid(rule):
     assert re.match(r"[A-Z]+\.[1-9][0-9]*\.[1-9][0-9]*", rule.identifier)
     assert rule.identifier.startswith(rule.parent.identifier)
 
 
-def test_rule_identifiers_strictly_increasing(subsection):
-    for index, rule in enumerate(subsection.rules):
-        assert rule.identifier.split(".")[-1] == str(index + 1)
+def test_rule_identifiers_strictly_increasing(enumerated_rule):
+    index, rule = enumerated_rule
+    assert rule.identifier.split(".")[-1] == str(index + 1)
 
 
 def test_rule_identifier_follows_convention(rule):
--- a/tests/test_convention_doc/test_rules.py
+++ b/tests/test_convention_doc/test_rules.py
@@ -2,8 +2,6 @@
 
 import re
 
-import pytest
-
 
 def test_rule_identifier_valid(rule):
     assert re.match(r"[A-Z]+\.[1-9][0-9]*\.[1-9][0-9]*", rule.identifier)
--- a/tests/test_convention_doc/test_rules.py
+++ b/tests/test_convention_doc/test_rules.py
@@ -1,0 +1,27 @@
+import re
+
+import pytest
+
+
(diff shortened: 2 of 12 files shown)


tests/test_convention_doc/test_rules.py
```diff
@@ -0,0 +1,27 @@
+"""Tests for the convention rules"""
+
+import re
+
+
+def test_rule_identifier_valid(rule):
+    assert re.match(r"[A-Z]+\.[1-9][0-9]*\.[1-9][0-9]*", rule.identifier)
```
mshafer-NI, line 7: this could be shortened with:
```
r"[A-Z]+(\.[1-9][0-9]*){2}"
```

(but I'm not sure it reads better), although it might..

Alternatively, define these in a shared file (so also used by the other test):
```
SECTION_RE = r'[A-Z]+'
RULE_RE = r'[1-9][0-9]*'
```
Then this becomes:
```python
assert re.match(f'{SECTION_RE}\\.{RULE_RE}\\.{RULE_RE}")
```
(or)
```python
assert re.match(r'{}\.{}\.{}'.format(SECTION_RE, RULE_RE, RULE_RE))
```
```

File: tests/test_convention_doc/test_rules.py [hunk 2]
```diff
@@ -0,0 +1,27 @@
...
+    assert rule.identifier.startswith(rule.parent.identifier)
+
+
+def test_rule_identifiers_strictly_increasing(subsection):
+    for index, rule in enumerate(subsection.rules):
+        assert rule.identifier.split(".")[-1] == str(index + 1)
```
mshafer-NI, line 13: Clean Code principle: "Test should have only one reason to fail".

This test can fail on the first iteration, or the third, or the nth.

Eitehr - make the subsection.rules an index a fixture, or, use [`pytest-subtests`](https://pypi.org/project/pytest-subtests/#usage) to make this act as `n` tests

tests/test_convention_doc/test_rules.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+def test_rule_identifier_follows_convention(rule):
+    assert (
```
mshafer-NI, line 17:
```python
assert any([
...
])
```
??

tests/test_convention_doc/test_sections.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+def test_section_identifier_valid(section):
+    assert re.match(r"[A-Z]+", section.identifier)
+
+
+def test_section_identifiers_unique(sections):
+    section_identifiers = set()
+    for section in sections:
+        assert section.identifier not in section_identifiers
```
mshafer-NI, line 13: Clean Code principle: "Test should have only one reason to fail".

This test can fail on the first iteration, or the third, or the nth.

Either - make the sections fixture yield multiple, or, use [`pytest-subtests`](https://pypi.org/project/pytest-subtests/#usage) to make this act as n tests

tests/test_convention_doc/test_subsections.py
```diff
@@ -0,0 +1,23 @@
+"""Tests for the convention subsections"""
+
+import re
+
+
+def test_subsection_identifier_valid(subsection):
+    assert re.match(r"[A-Z]+\.[1-9][0-9]*", subsection.identifier)
```
mshafer-NI, line 7: more ugly regex strings - pretty sure I saw this one before . . .

tests/test_convention_doc/test_subsections.py
```diff
@@ -0,0 +1,23 @@
...
+    assert subsection.identifier.startswith(subsection.parent.identifier)
+
+
+def test_subsection_identifiers_strictly_increasing(section):
+    for index, subsection in enumerate(section.subsections):
+        assert subsection.identifier.split(".")[-1] == str(index + 1)
```
mshafer-NI, line 13: "A test should have only one reason to fail" - see above

tests/test_convention_doc/test_subsections.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+def test_subsection_isnt_rule(subsection):
+    assert not (
```
mshafer-NI, line 17:
```python
assert none([
...
])
```
??

tests/conftest.py
```diff
@@ -0,0 +1,37 @@
+import pytest
```
mshafer-NI, line 1: Should we be using a given, when, then structure for the test names??

tests/test_convention_doc/test_rules.py
thejcannon: I'll see what I can do. FWIW I think `pytest-subtests` kinda subverts `pytests` preference of knowing the test cases up front.

tests/test_convention_doc/test_rules.py
thejcannon: I feel like another file just for two trivial regex bits is a bit much... 

This doesn't worry me much as I like the regexs within the isolated tests, and they aren't much nastier than any other regexes. 🤷‍♂️ 

I'll leave this open to see what others think.

tests/test_convention_doc/test_rules.py
thejcannon: Okie dokie, fixed this by using another fixture which gives the enumerated results.

tests/test_convention_doc/test_rules.py
thejcannon: There's pros/cons of each, essentially do we care about the repetition or about indentation

tests/test_convention_doc/test_sections.py
thejcannon: Okie dokie, fixed this by using another fixture which gives the enumerated results.

tests/test_convention_doc/test_subsections.py
thejcannon: Okie dokie, fixed this by using another fixture which gives the enumerated results.

tests/conftest.py
thejcannon: I don't think we have a convention for that, really. I'd like to stick to stick to simple names until we agree on something more complex.

tests/test_convention_doc/test_rules.py
neilvana: I don't have any concerns about the regexes like this.  They aren't overly complicated.

PR #15, opened 2020-08-13
Add PEP 257 conventions

thejcannon:


docs/Coding-Conventions.md
```diff
@@ -813,18 +813,225 @@ import sys
...
+
+> 🐍 This rule stems from [PEP 257](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings)
+
+E.g. "Do this", "Return that" instead of "Returns the ...".
+
+### [D.1.11] ✔️ **DO** Start multiline docstrings with a one-line summary followed by a blank line
+
+> 🐍 This rule stems from [PEP 257](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings)
+
+The summary line may be on the same line as the opening quotes or on the next line.
```
thejcannon, line 948: I personally would like us to pick one style and stick to it with-regards-to summary line on triple-quote line or on the next line.

I'm partial to summary-on-next line. What do others think?

DavidCurtiss: Elsewhere we have said that we would follow google's docstring conventions, which have it on the same line as the opening quotes.

mshafer1: black enforces same line if a single line, but different line if multi-line.

thejcannon:
@mshafer1 I fully expect that PR to go in before this one, as this one is conventions doc changes which warrants many more eyes. 😄

docs/Coding-Conventions.md
thejcannon: @mshafer1 [I don't see that behavior](https://black.now.sh/?version=stable&state=_Td6WFoAAATm1rRGAgAhARYAAAB0L-Wj4AEOAHVdAD2IimZxl1N_WlXnON2nzNX9G_wLs0bF-u_tpYJovzJbQz9x9q03oS_HMK3ji0DEn6xPpYM6EeSY6FL50if9iIzJ9yIw3Zz3EXmLpndgg6IOjRughHZUEh5E8SduDgMyn95qNMbmNWBbh9tNniBbd2lclpK1bQAAAADjyMPBHz_e-wABkQGPAgAAiEqxnrHEZ_sCAAAAAARZWg==)

thejcannon: @DavidCurtiss I don't see this exact guidance on Google's Python Styleguide: https://google.github.io/styleguide/pyguide.html

(Although their docstrings are on the same line, they don't call it out)

DavidCurtiss: The flake8-docstrings module includes it as part of its google conventions:
```
> flake8 --docstring-convention=google t.py
systemlink\clients\tag\t.py:9:1: D212 Multi-line docstring summary should start at the first line
```

thejcannon: Ohhh interesting...

thejcannon:
Spamming everyone, since it is a convention change. Please [approve/review the PR](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/approving-a-pull-request-with-required-reviews). (If you have no opinion, then just approve)
@davidcurtiss, @alejandro5042, @irwand, @kcuzner, @jryckman, @neilvana, @mshafer-NI, @rick-morrow,@rtzoeller, @alexweav, @sbethur

docs/Coding-Conventions.md
thejcannon: I suppose we can open this can of worms then when we evaluate `flake8-docstrings` 😄

DavidCurtiss:
👍

docs/Coding-Conventions.md
```diff
@@ -813,18 +814,225 @@ import sys
...
+# Good
+class ArgumentClinic(object):
+    """An argument clinic which offers several courses.
+
+    Courses can be purchased in 5-minute or full half-hour intervals.
+    """
+
+    def argue(self):
+        """
+        Argue for the purchased amount of time.
```
irwand, line 869: Hmm.. I thought we always want to put 1st sentence/title on the same line always ?

thejcannon: PEP 8 says either is fine (I'm guessing because the trimming/dedenting will result in the same result).

I started a discussion above, we can go ahead and make the statement now, or wait until we can enforce something.

Does anyone have any strong opinions either for/against?

DavidCurtiss: I want to be able to rely on flake8-docstrings (or equivalent). But I'm fine waiting a little while until you address flake8-docstrings directly.

docs/Coding-Conventions.md
```diff
@@ -813,18 +814,251 @@ import sys
...
+
+> 🐍 This rule stems from [PEP 257](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings)
+
+```python
+# Bad
+class ArgumentClinic(object):
+    """An argument clinic which offers several courses."""
+    def argue(self):
+        pass
+
```
alejandro5042, line 952: Isn't there another rule that requires 2 empty lines between definitions?

docs/Coding-Conventions.md
```diff
@@ -813,18 +814,251 @@ import sys
 
 ## [D.1] Docstrings
 
-### [D.1.1] ✔️ **DO** Write docstrings for all public modules, functions, classes, and methods
+### [D.1.1] ✔️ **DO** Write docstrings for all public packages, modules, functions, classes, and methods
```
alejandro5042, line 817: Variables too?

docs/Coding-Conventions.md
```diff
@@ -813,18 +814,251 @@ import sys
...
 > 🐍 This rule stems from [PEP 8](https://www.python.org/dev/peps/pep-0008) and [PEP 257](https://www.python.org/dev/peps/pep-0257/#one-line-docstrings)
 
-### [D.1.3] ✔️ **DO** Put closing `"""` on its own line for multiline docstrings
+### [D.1.4] ✔️ **DO** Put closing `"""` on its own line for multiline docstrings
 
 > 🐍 This rule stems from [PEP 8](https://www.python.org/dev/peps/pep-0008) and [PEP 257](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings)
 
+```python
+# Bad
+class ArgumentClinic(object):
```
alejandro5042, line 845: The word "argument" is overloaded in this context (e.g. in terms of a function argument or parameter).

Suggestion: Use concepts or words from simple, everyday life to avoid confusion.

Besides, what is an "argument clinic" anyway?

docs/Coding-Conventions.md
```diff
@@ -813,18 +814,251 @@ import sys
...
+
+    This function will throw an OutOfTimeException when the time has elapsed.
+    """
+```
+
+### [D.1.12] ✔️ **DO** Format a script's docstring as its "usage" message
+
+> 🐍 This rule stems from [PEP 257](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings)
+
+The docstring should be printed when the script is invoked with incorrect or missing arguments (or perhaps with a "-h" option, for "help").
```
alejandro5042, line 1003: Use backticks for technical/code terms, such as `-h`.

docs/Coding-Conventions.md
```diff
@@ -813,18 +814,251 @@ import sys
...
+It should document the script's function and command line syntax, environment variables, and files.
+The docstring can be fairly elaborate (several screens full) and should be sufficient for a new user to use the command properly,
```
alejandro5042, line 1005: I don't think we should encourage "several screens full" of usage strings. Few people will read/appreciate such verbose documentation. We should encourage precise, comprehensive, and concise documentation... the shorter the better.

docs/Coding-Conventions.md
```diff
@@ -813,18 +814,251 @@ import sys
...
+    This function will throw an OutOfTimeException when the time has elapsed.
+    """
+```
+
+### [D.1.12] ✔️ **DO** Format a script's docstring as its "usage" message
+
+> 🐍 This rule stems from [PEP 257](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings)
+
+The docstring should be printed when the script is invoked with incorrect or missing arguments (or perhaps with a "-h" option, for "help").
+It should document the script's function and command line syntax, environment variables, and files.
```
alejandro5042, line 1004: "Files" could be misinterpreted to mean the script files. Instead, consider wording to "configuration/input/output files" or similar.

docs/Coding-Conventions.md
```diff
@@ -813,18 +814,251 @@ import sys
...
+
+This includes (if applicable) the class's:
+
+- overall behavior
+- public methods
+- instance variables
+- additional info for subclasses
+
+It should not include the specific documentation for the constructor or methods.
+
```
alejandro5042, line 1042: Did you mean to have a double blank line here?

docs/Coding-Conventions.md
```diff
@@ -813,18 +814,251 @@ import sys
...
+    """Argue for the purchased amount of time.
+
+    This function will throw an OutOfTimeException when the time has elapsed.
+    """
+
+def argue2(minutes):
+    """
+    Argue for the purchased amount of time.
+
+    This function will throw an OutOfTimeException when the time has elapsed.
```
alejandro5042, line 995: Is there a format language we should use for docstrings? Markdown? Or just plain text?

alejandro5042, line 995: Do we have a rule mentioning complete, grammatically correct sentences? (ending with a period, for example)

docs/Coding-Conventions.md
```diff
@@ -813,18 +814,251 @@ import sys
...
+
+### [D.1.16] ✔️ **DO** Fully document a class in its docstring
+
+> 🐍 This rule stems from [PEP 257](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings)
+
+This includes (if applicable) the class's:
+
+- overall behavior
+- public methods
+- instance variables
```
alejandro5042, line 1038: *Public* instance variables?

docs/Coding-Conventions.md
```diff
@@ -813,18 +814,251 @@ import sys
...
+
+### [D.1.13] ✔️ **DO** List exported modules and subpackages in a package's docstring
+
+> 🐍 This rule stems from [PEP 257](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings)
+
+### [D.1.14] ✔️ **DO** List exported objects (classes, functions, exceptions, etc...) in a module's docstring
+
+> 🐍 This rule stems from [PEP 257](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings)
+
+Each documented object should have a one-line summary (with less detail than the summary line of the objects' docstring)
```
alejandro5042, line 1016: PEP 257 says:

> The docstring for a module should *generally* list the classes, exceptions and functions (and any other objects) that are exported by the module, with a one-line summary of each.

Emphasis mine.

Are we being stricter than we need to be here? To list every documented object in the module docstring seems like a lot of work for little gain, especially for larger modules. Seems like the module should just describe itself and its major components.

docs/Coding-Conventions.md
DavidCurtiss: https://github.com/ni/python-styleguide/blob/master/docs/Coding-Conventions.md#f22-%EF%B8%8F-do-surround-top-level-function-and-class-definitions-with-two-blank-lines

thejcannon: That's _between_ definitions. The class docstring isn't a "definition" per-se.

docs/Coding-Conventions.md
thejcannon: Unspecified at the moment 😄. I believe when we cross the documentation bridge, the convention is RST, but don't quote me. Right not we're just saying "have one" and I guess the implicit assumption is plain text.

thejcannon: `D.1.9` is "End your docstrings with a period", but I don't think we have one about using complete sentences. I think it's worth adding.

docs/Coding-Conventions.md
thejcannon: I agree (in my defense I copied this form PEP 257 😄)

docs/Coding-Conventions.md
thejcannon: So this is specifically about the usage string of a script, which I think is OK. I think it should be the shortest amount of information possible to cover everything the caller needs to know (and maybe some extra). 

What I usually see is either no documentation (ew) or not enough, forcing me to go find what the options mean/do.

That being said, I'd be OK with just removing the "several screens full" as the sentence without it captures what we both want (I think)

>The docstring can be fairly elaborate and should be sufficient for a new user to use the command properly,

docs/Coding-Conventions.md
thejcannon: Would changing this to a "CONSIDER" be sufficient?

docs/Coding-Conventions.md
kcuzner: https://www.youtube.com/watch?v=xpAvcGcEc0k

thejcannon: In keeping with Python's culture every single example (so far, and if I get my way always) has come from some Monty Python production. 😁

thejcannon: I'm open to suggestions for improvement, though!

DavidCurtiss: No you aren't.

alejandro5042: I dunno then, maybe use anything except the word "Argument", which may only confuse readers (including myself).

I'm all for fun and emojis and funny references; don't get me wrong! Maybe pick another fun reference from the movies that doesn't use an overloaded programming term (argument) that doesn't require you to squint at or require knowledge of history behind Python and its connection to Monty Python. Coding conventions are already complicated by nature, and should be accessible to everyone -- especially newcomers -- globally and in every culture :)

That said, take this comment as a _light_ suggestion. I am OK without a change.

docs/Coding-Conventions.md
alejandro5042: "elaborate" is maybe the word I'm getting stuck on 😋 

When I think "elaborate," I think pages of blocky, intense documentation; e.g. Git. Everyone knows Git is not an example of newcomer-helpful documentation. It is *correct* however and extremely useful. I think we can aim for a helpful balance.

> The docstring should be comprehensive yet concise enough for new users to get going quickly and correctly. Use sections to delineate essential details (such as what the command does) from remarks (such as edge cases). Link to further documentation, if necessary.

Thoughts?

docs/Coding-Conventions.md
alejandro5042: Great! Maybe we can adapt D.1.9 to be that way? Up to you.

docs/Coding-Conventions.md
alejandro5042: Well you probably SHOULD highlight your most popular or core APIs in the module docstring.

For example, a `time` library may have `DateTime` and `Duration` classes. But it's not super important to mention `Timezone` or `GregorianCalendar`.

> A module for time-related operations, providing classes for `DateTime` and `Duration` and associated functionality (string parsing, timezones, daylight savings, etc).

docs/Coding-Conventions.md
alejandro5042: Not sure if I follow. The rule says:

> [F.2.2] ✔️ DO Surround top-level function **and class definitions** with two blank lines
> 🐍 This rule stems from PEP 8

Emphasis mine.

Do we need to update the rule? But if we do, it seems arbitrary.

DavidCurtiss: Are you asking for an extra blank line between the "Bad" class and the "Good" class in the example?

alejandro5042: Yes. Anytime we have definitions we should be using double blank lines, as noted by the current wording of F.2.2. And we should follow (or amend) our coding conventions in our coding conventions document ;)

docs/Coding-Conventions.md
mshafer1: public (global) variables are a no-no in Python

thejcannon: They are certainly frowned upon, but not unheard-of (See the different flags in `re`).

But the more important thing to point out is you can't mechanically document a variable like you can the other things 😄

docs/Coding-Conventions.md
thejcannon: I suppose I'll just take this time to watch more Monty Python at work to find better terms to use 😉.

docs/Coding-Conventions.md
thejcannon: The more I think about it the more I'm actually starting to disagree with the convention altogether. In the modern age, libraries are used to handle script arguments and there's two flavors: `docopt`-style in which case this convention is redundant, and `argparse`-style in which case the help message is programmatically generated.

The last thing I'd want is someone copy `argparse` output into their docstring!

(Not to mention [they don't even follow the guideline themselves](https://github.com/python/cpython/blob/master/Lib/http/server.py), hmpf

What do people think about removing this one?

docs/Coding-Conventions.md
thejcannon: Yeah let's try it out 😄

docs/Coding-Conventions.md
thejcannon: Ah, this one's on me. When `git` had the little code preview I only saw line 954 and onward. I agree 😄

docs/Coding-Conventions.md
mshafer-NI: First reaction - I didn't know about this, I've only ever used the built-in "help()" method.

Note:
the community has built up tools to compile doc strings and method signatures into your public api documentation (e.g., readthedocs)

IFF you're building a CLI - the argument/options/effect of the program should all be documented through argparse such that running "-h" acts like a normal cli and gives you a list of arguments and what they do (which is the default argparse behavior as long as you provide `help="..."` on your arguments)

IF you're building a utility module - something that is intended to be imported, you probably wouldn't provide argparse. - (I have seen people have argparse get it's help string via *class*`.__doc__`, but this is generally not good practice as it is using the classes internal "magic" propery.)

In the latter case, if a docstring is provided for the module that does "document the [module's] function and [public api entry points]" - then running
```python
import x
help(x)
```
in the interpreter *should* yield helpful information about the public api and how to use it.

Example:
```
>>> import os
>>> help(os)
Help on module os:

NAME
    os - OS routines for NT or Posix depending on what system we're on.

DESCRIPTION
    This exports:
      - all functions from posix or nt, e.g. unlink, stat, etc.
      - os.path is either posixpath or ntpath
      - os.name is either 'posix' or 'nt'
      - os.curdir is a string representing the current directory (always '.')
      - os.pardir is a string representing the parent directory (always '..')
      - os.sep is the (or a most common) pathname separator ('/' or '\\')
      - os.extsep is the extension separator (always '.')
      - os.altsep is the alternate pathname separator (None or '/')
      - os.pathsep is the component separator used in $PATH etc
      - os.linesep is the line separator in text files ('\r' or '\n' or '\r\n')
      - os.defpath is the default search path for executables
      - os.devnull is the file path of the null device ('/dev/null', etc.)

    Programs that import and use 'os' stand a better chance of being
    portable between different platforms.  Of course, they must then
    only use functions that are defined by all platforms (e.g., unlink
    and opendir), and leave all pathname manipulation to os.path
    (e.g., split and join).

CLASSES
    builtins.Exception(builtins.BaseException)
        builtins.OSError
...
```


Suggested alteration:

```
The docstring should be printed when the module "help" is called on the module.
It should document the module's function and top level API, environment variables, and files.
The docstring can be fairly elaborate (several screens full) and should be sufficient for a new user to use the module properly.
```

docs/Coding-Conventions.md
thejcannon: I'm not good at wordsmithing, how's this?

> List __the most relevant__ exported objects (classes, functions, exceptions, etc...) in a module's docstring

docs/Coding-Conventions.md
thejcannon: I changed it to use references from the cheese shop sketch

docs/Coding-Conventions.md
mshafer-NI: hmmmm. . . so, remove this? or encourage people to pass the docstring to argparse as the description??

(in favor of remove)

thejcannon: Removed.

thejcannon:
I'd like to get this in soon.
@mshafer-NI @ScotSalmon @alejandro5042 Any more suggestions?

docs/Coding-Conventions.md
```diff
@@ -813,18 +814,234 @@ import sys
...
 
-### [D.1.2] ✔️ **DO** Put closing `"""` on the same line for one line docstrings
+ℹ️ You can document a package by documenting the module docstring of the package directory's `__init__.py`
+
+### [D.1.2] ✔️ **DO** Use triple double quotes for docstrings
+
+> 🐍 This rule stems from [PEP 257](https://www.python.org/dev/peps/pep-0257/#one-line-docstrings)
+
+```python
+# Bad
```
mshafer-NI, line 828: also bad: `"Lumberjack: Cuts down trees, among other things"`

(since we're explicitly saying triple double-quotes, maybe show an example of both ways this can be violated)

thejcannon: Good idea!

docs/Coding-Conventions.md
alejandro5042: Sounds good!

docs/Coding-Conventions.md
alejandro5042: Great! LGTM

PR #16, opened 2020-08-18
Replace CODEOWNERS with a naggy bot

thejcannon:


thejcannon:
FYI @abaretto I gave up on the validation 😬

ScotSalmon:
Bummer.

alejandro5042:
> FYI @abaretto I gave up on the validation 😬

By the way, I didn't get this since that's not my username ;) Who knows who you at-mentioned!

It's `@alejandro5042`.

thejcannon:
Lol oops. At least it's right in the bot 😄

thejcannon:
> Can we have developers pay for GitHub?

We can, but I heard from more than one dev they didn't feel the cost was justified for only this use-case. I don't think it's out of the question, but since it isn't the immediate solution I want to make sure we're able to execute _today_.

alejandro5042:
@benmont is working to see how we can obtain licenses org-wide. Hopefully we can use the built-in tools as opposed to maintaining workarounds (although this is a pretty awesome workaround).

PR #18, opened 2020-08-26
Add our line length convention

thejcannon:
This is the result of the discussion(s)/forms we have had along with the excellent presentation by @mshafer-NI.


docs/Coding-Conventions.md
```diff
@@ -52,7 +52,38 @@ In all cases where a convention comes from a PEP, it will be marked as such.
...
+### [F.1.1] ✔️ **DO** Limit your lines to a maximum length of 100 characters
+
+There is no one-size-fits-all when it comes to a maximum line length. Too short and
+developers start contorting their code to fit the restriction. Too long and lines exceed
+what is visible in most common tools (code editors, diff visualizers, etc...). Additionally,
+automatic formatting can reduce the burden of maintaining a maximum line length, but can
+also be eager in enforcing it. In the end, choosing a maximum line length isn't about
+optimization, but is rather about finding a middle-ground that developers can agree on.
+
+We have chosen 100 characters because to some developers 80/88 charcters is too limiting,
```
rtzoeller, line 64:
```suggestion
We have chosen 100 characters because to some developers 80/88 characters is too limiting,
```

thejcannon: Oh boy do I need to install a spellcheck VS Code extension...

docs/Coding-Conventions.md
```diff
@@ -52,7 +52,38 @@ In all cases where a convention comes from a PEP, it will be marked as such.
 
 ## [F.1] Indents
 
-### [F.1.1] ✔️ **DO** Use 4 spaces per indentation level (never tabs)
+### [F.1.1] ✔️ **DO** Limit your lines to a maximum length of 100 characters
```
mshafer-NI, line 55: this is more of a note for future, adding rules into an order like this will effectively break perma-links (that is, if someone has commented on a PR that "to comply with F.1.1 (with a link), we should use spaces", after this goes in, that link won't make any sense.

This is probably fine now because we're before much of an adoption at all, but in the future, this could become a drastic problem.

thejcannon: Yes. As long as this is a work in progress I fully accept that identifiers might change. I actually think it's going to be desirable, as the section/subsection identifiers will surely shuffle once we are able to see "the big picture" of our conventions.

Once we "go public", identifiers __will never__ change (for the reason you give).

thejcannon:
@mshafer-NI  @DavidCurtiss  @rtzoeller  sorry for the noise. Going to close this one in favor of #20, which spammed the group members correctly 🎉

PR #21, opened 2020-08-26
Prepare the styleguide for automated convention testing

thejcannon:
This PR prepares the styleguide (both the document and the tooling) for having `pytest` enforce that our code examples in the convention are honest with regards to the tool. Overview of the changes:

- Already merged in https://github.com/ni/python-styleguide/pull/20
- Bad/Good codeblocks now split such that each codeblock is one of Bad/Good/Best
- Enforces we document which conventions are auto-enforced through the examples
- Enforces all Good/Best blocks cause no lint errors
- Enforces we document the error codes a bad codeblock represents

(An example of me testing this out with `pycodestyle` can be found on my branch: https://github.com/thejcannon/python-styleguide/blob/pycodestyle/docs/Coding-Conventions.md. Please note the link won't be valid forever as the branch will eventually be deleted)

---

I also want to outline here "the big picture" and some of _why_ I'm leaning this way, in case it has been lost to the wind.

- I want us to leverage as many lint plugins as makes sense. Tools are stronger than humans here.
- When we add a lint plugin, I want us to evaluate the impact it will have on our conventions (by also changing the conventions)
- In order to help all of us evaluate _ideally_ all the possible codes a plugin raises, an exhaustive set of examples can be produced in the form of codeblocks in conventions 

So my hope is that this will be a good way to help us achieve the above, by letting the tooling to the talking.

### DIFF
--- a/tests/test_convention_doc/conftest.py
+++ b/tests/test_convention_doc/conftest.py
@@ -45,3 +45,54 @@ def subsection(request):
 )
 def rule(request):
     return request.param
+
+
+@pytest.fixture(
+    scope="module",
+    params=[
+        pytest.param(codeblock, id=f"{rule.identifier}-{codeblock.descriptor}")
+        for rule in doctypes.RULES
+        for codeblock in rule.codeblocks
+    ],
+)
+def codeblock(request):
+    return request.param
+
+
+@pytest.fixture(
+    scope="module",
+    params=[
+        pytest.param(codeblock, id=f"{rule.identifier}-{codeblock.descriptor}")
+        for rule in doctypes.RULES
+        for codeblock in rule.codeblocks
+        if codeblock.descriptor == "Bad"
+    ],
+)
+def bad_codeblock(request):
+    return request.param
+
+
+@pytest.fixture(
+    scope="module",
+    params=[
+        pytest.param(codeblock, id=f"{rule.identifier}-{codeblock.descriptor}")
+        for rule in doctypes.RULES
+        for codeblock in rule.codeblocks
+        if codeblock.descriptor == "Good"
+    ],
+)
+def good_codeblock(request):
+    return request.param
+
+
+@pytest.fixture(
+    scope="module",
+    params=[
+        pytest.param(codeblock, id=f"{rule.identifier}-{codeblock.descriptor}")
+        for rule in doctypes.RULES
+        for codeblock in rule.codeblocks
+        if codeblock.descriptor == "Best"
+    ],
+)
+def best_codeblock(request):
+    return request.param
--- a/tests/test_convention_doc/doctypes.py
+++ b/tests/test_convention_doc/doctypes.py
@@ -46,13 +46,36 @@ class Rule(_RuleHeader):
         self.is_automatically_enforced = self.header_text.endswith("💻")
         self.is_automatically_fixed = self.header_text.endswith("✨")
         self.codeblocks = [
-            Codeblock(self, match["language"], match["description"], match["body"],)
-            for match in re.finditer(
-                r"```(?P<language>[a-z]+\n)?(# (?P<description>.*?)\n)?(?P<body>.*?)```",
-                self.body_text,
-                flags=re.DOTALL,
-            )
+            Codeblock(self, match["language"], match["description"], match["body"])
+            for match in Rule._find_codeblocks(self.body_text)
         ]
+
+    @staticmethod
+    def _find_codeblocks(text):
+        r"""Find codeblocks in given text, yielding match objects.
+
+        >>> [match.groupdict() for match in Rule._find_codeblocks('```x=5```')]
+        [{'language': None, 'description': None, 'body': 'x=5'}]
+
+        >>> [match.groupdict() for match in Rule._find_codeblocks('```python\nx=5```')]
+        [{'language': 'python', 'description': None, 'body': 'x=5'}]
+
+        >>> [
+        ...     match.groupdict()
+        ...     for match in
+        ...     Rule._find_codeblocks('```python\n# Bad\nx=5```')
+        ... ]
+        [{'language': 'python', 'description': 'Bad', 'body': 'x=5'}]
+
+        >>> [match.groupdict() for match in Rule._find_codeblocks('```\n# Bad\nx=5```')]
+        [{'language': None, 'description': 'Bad', 'body': 'x=5'}]
+
+        """
+        yield from re.finditer(
+            r"```((?P<language>[a-z]+)\n)?(\n?# (?P<description>.*?)\n)?(?P<body>.*?)```",
+            text,
+            flags=re.DOTALL,
+        )
 
     @property
     def error_codes(self):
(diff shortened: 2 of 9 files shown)


tests/test_convention_doc/test_codeblocks.py
```diff
@@ -0,0 +1,46 @@
...
+def test_bad_codeblocks_document_lint_errors(lint_codeblock, bad_codeblock, capsys):
+    if bad_codeblock.rule.is_automatically_enforced:
+        assert not lint_codeblock(
+            bad_codeblock, "--format", "'%(code)s'"
+        ), 'Expected automatically enforced "bad" codeblock to cause lint error'
+
+        captured = capsys.readouterr()
+        error_codes = set(code.strip("'") for code in captured.out.strip().split("\n"))
+
+        assert error_codes.issubset(
```
mshafer-NI, line 31: ummm... if error_codes is only a subset of bad_codebock.rule.error_codes...

That would allow for not every error code in the rule to be reported in this case?

(is this intentional, thinking that we should have an example per error code? Do we have a check that we have them all covered?)

tests/test_convention_doc/test_sections.py
```diff
@@ -17,5 +17,4 @@ def test_section_identifiers_unique(sections):
 
 
 def test_section_identifier_follows_case_convention(section):
-    section_title = ("**".join(section.header_text.split("**")[2:])).strip()
-    assert section_title[0] == section_title[0].upper()
+    assert section.header_text[0] == section.header_text[0].upper()
```
mshafer-NI, line 20: `assert section.header_text.isupper()`

tests/test_convention_doc/test_subsections.py
```diff
@@ -39,5 +39,4 @@ def test_subsection_isnt_rule(subsection):
 
 
 def test_subsection_identifier_follows_case_convention(subsection):
-    subsection_title = ("**".join(subsection.header_text.split("**")[2:])).strip()
-    assert subsection_title[0] == subsection_title[0].upper()
+    assert subsection.header_text[0] == subsection.header_text[0].upper()
```
mshafer-NI, line 42: `assert subsection.header_text[0].isupper(), "header text should be 'ALL CAPS'"

tests/test_convention_doc/conftest.py
```diff
@@ -45,3 +45,54 @@ def subsection(request):
...
+
+
+@pytest.fixture(
+    scope="module",
+    params=[
+        pytest.param(codeblock, id=f"{rule.identifier}-{codeblock.descriptor}")
+        for rule in doctypes.RULES
+        for codeblock in rule.codeblocks
+        if codeblock.descriptor == "Bad"
+    ],
```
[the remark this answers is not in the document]
thejcannon: `pytest` isn't scared by this, it enumerates duplicate IDs automatically 🎉

tests/test_convention_doc/doctypes.py
```diff
@@ -43,9 +43,34 @@ class Rule(_RuleHeader):
     def __init__(self, text, **kwargs):
         super().__init__(text, **kwargs)
 
-        self.is_automatically_enforced = self.header_text.endswith("🖥️")
+        self.is_automatically_enforced = self.header_text.endswith("💻")
         self.is_automatically_fixed = self.header_text.endswith("✨")
```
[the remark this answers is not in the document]
thejcannon: The honest truth is I don't know yet. I suspect you are right that it would be additive, but we could also just maintain that fixed _implies_ enforced (if a tool is smart enough to fix it, then there's no reason we can't catch it). I think we'll cross the bridge when we get to it.

For this PR however, we aren't changing assumptions, nor should we.

tests/test_convention_doc/doctypes.py
```diff
@@ -43,9 +43,34 @@ class Rule(_RuleHeader):
...
-        self.codeblocks = re.findall(r"```(.*?)```", self.body_text)
+        self.codeblocks = [
+            Codeblock(self, match["language"], match["description"], match["body"],)
+            for match in re.finditer(
+                r"```(?P<language>[a-z]+\n)?(# (?P<description>.*?)\n)?(?P<body>.*?)```",
```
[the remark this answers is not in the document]
thejcannon: Not sure I follow, I don't expect us to have one-line codeblocks that we intend to lint, but maybe you have something in mind I don't?

tests/test_convention_doc/test_codeblocks.py
thejcannon: This came up when I was trying to give examples of the "indent 4 spaces and never tabs". I had an example of each, but each example in individuality didn't cause _all_ the errors, just a subset.

> Do we have a check that we have them all covered?

No. It's possible. Probably desirable. But it's not here. I'll wait to solicit feedback on the overall approach first.

tests/test_convention_doc/test_sections.py
thejcannon: This is was actually a _really_ good idea... turns out I was testing the leading space before 😱

tests/test_convention_doc/doctypes.py
```diff
@@ -43,9 +43,34 @@ class Rule(_RuleHeader):
...
+                self.body_text,
+                flags=re.DOTALL,
+            )
+        ]
+
```
mshafer-NI, line 56: RE: https://github.com/ni/python-styleguide/pull/21/files/8c48341eb68936ac1681640212c15b3d3b45d2c9#r478422560

we should either test or document such expectations.

(note, untested)

```suggestion
       self.codeblocks = [Codeblock(self, language, description, body,) for language, description, body in _RuleHeader.__find_code_blocks(self.body_text)]
        

    @staticmethod
    def __find_code_blocks(text):
    '''
    >>> list(_RuleHeader.__find_code_blocks("""\n```python\nx = 5\n```"""))
    [('python', '', 'x = 5']

    >>> list(_RuleHeader.__find_code_blocks("""\n```python\n# description\nx = 5\n```"""))
    [('python', '# description', 'x = 5']
    
    >>> list(_RuleHeader.__find_code_blocks("""```x = 5```"""))
    Traceback: most recent call last
     ...
    >>> list(_RuleHeader.__find_code_blocks("""```some other language\n# despcriont\nx = 5\n```"""))
    Traceback: most recent call last
     ...
    '''
    for match in re.finditer(
           r"```(?P<language>[a-z]+\n)?(# (?P<description>.*?)\n)?(?P<body>.*?)```",
           text,
           flags=re.DOTALL,
    ):
        yield match["language"], match["description"], match["body"]
```

thejcannon: Since this code gets run when tests are being collected, I'd strongly prefer if we didn't raise any exceptions. I'll work on this more, but this is also test code, so I don't think it needs to be perfect.

mshafer-NI: >this is also test code, so I don't think it needs to be perfect.

Strongly disagree with this statement - test code is the most important code.

thejcannon: Added some doctests

PR #28, opened 2020-09-25
Fix tests by adding format param

thejcannon:
The tests were passing because we were asserting the command failed, and then testing that the errors the command resulted in was a subset of what we expected. Since we didn't have the format param, the errors we produced was an empty set 😢.

This is fixed by fixing the format param issue and also asserting there are some error codes in the output.


mshafer-NI:
Why is the PR showing no checks??

thejcannon:
Fixed the workflow file: #29

mshafer-NI (approved):

irwand (approved):

kcuzner (approved):

rick-morrow (approved):

thejcannon:
Bypassing @neilvana, enough reviewers approved.

PR #30, opened 2020-10-15
Turn on pycodestyle linting

thejcannon:
`pycodestyle` is inlcluded by default in `flake8`, and handles some PEP8 conventions `black` doesn't format (as it doesn't pertain to style).

Additionally, most of our internal projects use the pycodestyle checks.


docs/Coding-Conventions.md
```diff
@@ -89,11 +93,13 @@ and to others 110/120 is too long.
 ```python
 # Bad
 line_with_101_chars = "spaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaam"
+# Also applies to looooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooong comments
```
rtzoeller, line 96: If I have a comment with a URL over 100 characters, what are my options?

mshafer-NI, line 96: shouldn't this be its own "bad" block so that both lines are throwing the error and are tested as such?

mshafer-NI: or is this the w505 so asserting both codes covers that?

mshafer-NI (approved):

docs/Coding-Conventions.md
thejcannon: Suffixing it with `noqa: W505` I believe is the only thing 😞 .

I would love for the tooling to recognize "this can't be split" and not error (like how `black` behaves), but unfortunately that's not the case today.

rtzoeller: It doesn't seem like that hard of a check to add a basic implementation for: if the comment is a single token, allow it to be longer than the minimum length.

E.g. your comment could pass by being rewritten to 

```
# Also applies to
# looooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooong
# comments
```

(assuming looo...ooong was greater than the maximum length).

rtzoeller: (Not suggesting you add support for this, just surprised it doesn't exist).

alejandro5042 (approved):

neilvana (approved):

rtzoeller (approved):

docs/Coding-Conventions.md
thejcannon: I completely agree with you. I've had a recent desire (_toootally unrelated_) for a better lint/format framework and have thought about this use case as URLs can get long as hell. No one cares about the contents so as long as you are sure there isn't anything after the URL (way over there past column 200).

rick-morrow (approved):

kcuzner (approved):

docs/Coding-Conventions.md
thejcannon: In the future, we outta change this to have the error(s) reported on a per-codeblock basis.

irwand (approved):

nethrasuresh (approved):

DavidCurtiss (approved):

docs/Coding-Conventions.md
```diff
@@ -175,6 +185,40 @@ Examples:
 - input\_
 - file\_
 
+### [N.1.3] ❌ **DO NOT** Use the characters 'l' (lowercase letter el), 'O' (uppercase letter oh), or 'I' (uppercase letter eye) as single character variable names 💻
+
+> 🐍 This rule stems from [PEP 8](https://www.python.org/dev/peps/pep-0008)
```
pakdev, line 190: Could link to the actual location: https://www.python.org/dev/peps/pep-0008/#names-to-avoid

thejcannon: We could, but then we'd probably want to link the others to be consistent. The reason I didn't do that initially was because it was tedious and didn't add any value. All this is meant to do is let the reader know this convention aligns strongly with the community.
We could honestly remove the hyperlinks altogether so these are just text all the way. I'll probably do that in a follow-up PR.

ni_python_styleguide/_cli.py
```diff
@@ -101,25 +98,16 @@ def main(ctx, verbose, quiet, config, exclude, extend_exclude):
...
-                # [tool.black] setting (which makes sense if you think about it)
-                # So we need to give it one
-                f"--black-config={str((pathlib.Path(__file__).parent / 'config.toml').resolve())}",
-                *file_or_dir,
-            ],
-        )
-    )
+    args = [
+        _qs_or_vs(obj["VERBOSITY"]),
+        f"--config={str((pathlib.Path(__file__).parent / 'config.ini').resolve())}",
```
pakdev, line 103: You don't have to call `str()` in f-strings, right?

ni_python_styleguide/_cli.py
```diff
@@ -101,25 +98,16 @@ def main(ctx, verbose, quiet, config, exclude, extend_exclude):
...
+        f"--exclude={obj['EXCLUDE']}" if obj["EXCLUDE"] else "",
+        f"--format={format}" if format else "",
+        # The only way to configure flake8-black's line length is through a pyproject.toml's
+        # [tool.black] setting (which makes sense if you think about it)
+        # So we need to give it one
+        f"--black-config={str((pathlib.Path(__file__).parent / 'config.toml').resolve())}",
+        *file_or_dir,
+    ]
+    app.run(list(filter(bool, args)))
```
pakdev, line 112: I don't see any guidance in our conventions about avoiding `filter` and `map`, but isn't Guido generally against their usage (esp in this case where we enumerate the iterable)?

pakdev (approved):

docs/Coding-Conventions.md
```diff
@@ -42,6 +42,10 @@ In all cases where a convention comes from a PEP, it will be marked as such.
 - [PEP 8 -- Style Guide for Python Code](https://www.python.org/dev/peps/pep-0008/)
 - [PEP 257 -- Docstring Conventions](https://www.python.org/dev/peps/pep-0257/)
 
+# Python versions
+
+This document is applicable to all Python versions which are not end-of-life.
+
```
sbethur, line 48: Having this is fine, but I'm just curious why we need to restrict the conventions?

ni_python_styleguide/config.ini
```diff
@@ -0,0 +1,39 @@
+# This file exists for us to configure the various tools we execute through config files.
+# Some tools (like flake8-black) can only be configured through a config.
+# Other tools, we leverage the config since it allows us to comment on the values.
+
+[flake8]
+max-line-length = 100
+max-doc-length = 100
+
+ignore =
+    # pycodestyle checks that are either enforced by, or go against flake8-black
```
sbethur, line 10: How painful will it be to curate this list every time pycodestyle or black is updated?

ni_python_styleguide/config.ini
```diff
@@ -0,0 +1,39 @@
+# This file exists for us to configure the various tools we execute through config files.
+# Some tools (like flake8-black) can only be configured through a config.
+# Other tools, we leverage the config since it allows us to comment on the values.
+
+[flake8]
+max-line-length = 100
+max-doc-length = 100
```
sbethur, line 7: Can this be put under `ignore` list since black enforces them?

sbethur (approved):

ni_python_styleguide/_cli.py
thejcannon: 😅

ni_python_styleguide/_cli.py
thejcannon: I couldn't find a good consensus online regarding how to best accomplish this behavior.
I would totally agree on not using `map` and `filter` if the function argument was a lambda, but if you are using some object instead I'm more ambivalent. This particular usage though (`bool` filtering) I think gets _less_ explicit when you use a comprehension, but it isn't painfully obvious that you're filtering by booleanizing values:
`[arg for arg in args if arg]`. 

Ideally we'd write `[arg for arg in args if func(arg)]` or `[... if arg.func()]`, however using `bool(arg)` is equivalent to just `arg`, so we're stuck with some very uninteresting (albeit pirate-y) comprehension.

There's apparent `filter(None, args)` which does the same thing, but that's just wacky.

I'm going to punt until we discuss it further in the future.

docs/Coding-Conventions.md
thejcannon: It helps us focus discussions. 

For sure we're straight-up ignoring Python 2's existence, but this also means we don't have to think about older Python 3s.

ni_python_styleguide/config.ini
thejcannon: Less painful than when it was first written 😄.
Really the use-case is mostly to avoid duplicate errors, so if we get it wrong it isn't really fatal, just a bit noisier.

More realistically, I don't see `pycodestyle` really adopting any new rules and `black` has it's style mostly down.

ni_python_styleguide/config.ini
thejcannon: These are settings for the tool and not rules to be ignored, so not really no.

If you're asking if the rules which are configured through these settings can be ignored: also no 😄. `black` doesn't "enforce" a line length of X, so much as it says "when I can reformat things to fit line length X I will". The cases it can't break a long line up are usually long strings (and it won't break them up as it enforces it doesn't change the AST, which breaking up a string would be doing). That means that some lines of code will be past 100 chars and pass `black --check`.

Example:
```python
long_string = "I am a very very very very very very very very very very very very very long string"""

def function():
    """I am a very very very very very very very very very very very very very long docstring"""
```

jryckman (approved):

alexweav (approved):

stick152 (approved):

PR #31, opened 2020-10-23
Create GitHub Publish Action

rick-morrow:
Create workflow for publishing to PyPI.

### How to use this workflow
The idea is that we will contribute to the repository as usual until we are ready to publish a new package. We will then:

1. Bump the Poetry version (i.e. 0.1.0).
1. Create and push the corresponding annotated git tag, using GitHub convention of v+version (i.e. v0.1.0).

This workflow will key off the commit containing the tag to create a GitHub release (mostly just for organization and finding the tag, in the future we could put release notes here), and publish the package to PyPI.

### Notes
This is currently using my personal PyPI account to publish the package, but I've invited the opensource_ni account to be the owner.

https://pypi.org/project/ni-python-styleguide/ (I deleted the test package I published, but the location still exists).

I have an email out to Michael Phillips about transferring ownership of that account to another group to maintain, and to create an api key for it so we can use GitHub organization secrets to publish under our opensource_ni account.

For now, I think the current method is okay until we get the administrative work done.


.github/workflows/Publish-Package.yml
```diff
@@ -0,0 +1,57 @@
+name: Publish Package
+
+on:
+  push:
+    branches: [master]
+    tags:
+      - "v[0-9]+.[0-9]+.[0-9]+*" # Push events to matching v*, i.e. v1.0, v20.15.10
```
mshafer-NI, line 7: comment and regex don't match:
https://regex101.com/r/qZ5y8W/1

```suggestion
      - "v[0-9]+.[0-9]+*" # Push events to matching v*, i.e. v1.0, v20.15.10
```

mshafer-NI (approved):

.github/workflows/Publish-Package.yml
rick-morrow: I'm tempted to just remove the comment since it's an artifact from previous testing I forgot to remove. 

As far as the wildcard, it's meant to match the poetry version with a 'v' prepended. Could/would our poetry version ever just be major+minor without the patch? 

I tried to be able to pick up any semver we would reasonably want without having to resort to `- "v*"`

mshafer-NI: I'ld also be good with this:
(I doubt poetry will ever release without a patch number, as that wouldn't comply with standard python versioning)

```suggestion
      - "v[0-9]+.[0-9]+.[0-9]+*" # Push events to matching v*, i.e. v1.0, v20.15.10
      + "v[0-9]+.[0-9]+.[0-9]+*" # Push events to matching v*, i.e. v1.0.0, v20.15.10
```

thejcannon:
We should document the publishing steps in the CONTRIBUTING.md

thejcannon (commented):
>We will then:
> 1. Bump the Poetry version (i.e. 0.1.0).
> 2. Create and push the corresponding annotated git tag, using GitHub convention of v+version (i.e. v0.1.0).

Assuming the version in `pyproject.toml` is the "next" version, isn't that backwards? We'll be tagging the commit that just has the new version as the "release". 

I think we should be tagging the code first, _then_ bumping the number.

.github/workflows/Publish-Package.yml
```diff
@@ -0,0 +1,57 @@
+name: Publish Package
+
+on:
+  push:
+    branches: [master]
+    tags:
+      + "v[0-9]+.[0-9]+.[0-9]+*" # Push events to matching v*, i.e. v1.0.0, v20.15.10
```
thejcannon, line 7: Do you need to escape the dots?
```suggestion
      + "v[0-9]+\.[0-9]+\.[0-9]+*" # Push events to matching v*, i.e. v1.0.0, v20.15.10
```

.github/workflows/Publish-Package.yml
```diff
@@ -0,0 +1,57 @@
...
+    name: Publish Package
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: actions/setup-python@v2
+        with:
+          python-version: 3.6
+
+      - name: Set up Poetry cache for Python dependencies
```
thejcannon, line 40: I've never seen this before, what exactly is it caching?

.github/workflows/Publish-Package.yml
```diff
@@ -0,0 +1,57 @@
...
+          key: ${{ runner.os }}-poetry-${{ hashFiles('**/poetry.lock') }}
+          restore-keys: ${{ runner.os }}-poetry-
+
+      - name: Install Poetry
+        run: |
+          curl -fsS -o get-poetry.py https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py
+          python get-poetry.py -y
+          echo "$HOME/.poetry/bin" >> $GITHUB_PATH
+
+      - name: Install dependencies
```
thejcannon, line 53: Do we need to install the component's dependencies to build?

.github/workflows/Publish-Package.yml
```diff
@@ -0,0 +1,57 @@
...
+        uses: actions/create-release@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          tag_name: ${{ steps.tag.outputs.tag }}
+          release_name: ${{ steps.tag.outputs.tag }}
+          draft: false
+          prerelease: false
+
+  publish_package:
```
thejcannon, line 30: Why is this implemented as 2 jobs?

.github/workflows/Publish-Package.yml
```diff
@@ -0,0 +1,57 @@
...
+  create_release:
+    name: Create Release
+    runs-on: ubuntu-latest
+    steps:
+      - name: Get tag
+        id: tag
+        run: |
+          echo ::set-output name=tag::${GITHUB_REF#refs/tags/}
+
+      - name: Create Release
```
thejcannon, line 19: If any of the below steps fail, we'll still have a GitHub release, right?

I think this should probably come last.

.github/workflows/Publish-Package.yml
rick-morrow: `tags` uses wildcards to match where `.` is not a special character https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet

.github/workflows/Publish-Package.yml
mshafer-NI: the path below - the effect is when the worker spins up, it downloads the `~/.cache/pypoetry` that was cached off by poetry last time, (this is the dir that poetry puts venv's in - https://python-poetry.org/docs/configuration/)

The effect is calling `poetry install` will run:
if no dependency changes since last build: run through and see that all are up to date and proceed
if some are upgraded: upgrade those
if new/removed: handle those

It makes the `poetry install` and related commands go much faster.

.github/workflows/Publish-Package.yml
rick-morrow: Merged to one job and put step at the end

neilvana (approved):

irwand (approved):

rick-morrow:
I'm going to leave this PR up for now, but investigate alternate tagging/versioning strategies

PR #32, opened 2020-11-05
Turn on pyflakes linting

thejcannon:
This PR enables certain `pyflakes` checks with related changes.

Note the following checks are still "on" in the tool but not called out in the convention:

### F601 - MultiValueRepeatedKeyLiteral
```python
{'a': 1, 'a': 2}
```

### F602 - MultiValueRepeatedKeyVariable
```python
a = 1
{a: 1, a: 2}
```

### F631 - AssertTuple ("An assert of a non-empty tuple is always True")
```python
>>> assert (False, )
```

### F633 - InvalidPrintSyntax
```python
>>> print >>sys.stderr, "Hello"
Traceback (most recent call last):
File "<stdin>", line 1, in <module>
TypeError: unsupported operand type(s) for >>: 'builtin_function_or_method' and '_io.TextIOWrapper'. Did you mean "print(<message>, file=<output_stream>)"?
```

### F634 - IfTuple ("Booleanizing a non-empty tuple always is True")
```python
>>> if (False,):
...    print("F")
...
F
```

### F821 - UndefinedName (Just one example, use your imagination)
```python
>>> class foo:
...     foo
...
Traceback (most recent call last):
File "<stdin>", line 1, in <module>
File "<stdin>", line 2, in foo
NameError: name 'foo' is not defined
```

### F822 - UndefinedExport 
```python
__all__ = ["a"]
# But 'a' doesn't exist!
```

### F823 - UndefinedLocal
```python
>>> a = 1
>>> def fun():
...     a
...     a = 2
...     return a
...
>>> fun()
Traceback (most recent call last):
File "<stdin>", line 1, in <module>
File "<stdin>", line 2, in fun
UnboundLocalError: local variable 'a' referenced before assignment
```

### F901 - RaiseNotImplemented

('raise NotImplemented' should be 'raise NotImplementedError')

### F404 - LateFutureImport
```python
>>> import os
>>> from __future__ import print_statement
File "<stdin>", line 1
SyntaxError: future feature print_statement is not defined
```

### F406 - ImportStarNotPermitted
```python
>>> def a():
...     from os import *
...
File "<stdin>", line 1
SyntaxError: import * only allowed at module level
```

### F407 - FutureFeatureNotDefined
```python
>>> from __future__ import foobar
File "<stdin>", line 1
SyntaxError: future feature foobar is not defined
```

### F621 - TooManyExpressionsInStarredAssignment
```python
>>> a0, a1, ..., a255, *rest = range(1<<8 + 1)
File "<stdin>", line 1
SyntaxError: too many expressions in star-unpacking assignment
```

### F622 - TwoStarredExpressions
```python
>>> a, *b, *c = range(10)
File "<stdin>", line 1
SyntaxError: two starred expressions in assignment
```

### F701 - BreakOutsideLoop
```python
>>> break
File "<stdin>", line 1
SyntaxError: 'break' outside loop
```

### F702 - ContinueOutsideLoop
```python
>>> continue
File "<stdin>", line 1
SyntaxError: 'continue' not properly in loop
```

### F704 - YieldOutsideFunction
```python
>>> yield 4
File "<stdin>", line 1
SyntaxError: 'yield' outside function
```

### F706 - ReturnOutsideFunction
```python
>>> return
File "<stdin>", line 1
SyntaxError: 'return' outside function
```

### F707 - DefaultExceptNotLast
```python
>>> try:
...     pass
... except:
...     pass
... except ValueError:
...     pass
...
File "<stdin>", line 2
SyntaxError: default 'except:' must be last
```

### F831 - DuplicateArgument
```python
>>> def foo(bar, bar): pass
...
File "<stdin>", line 1
SyntaxError: duplicate argument 'bar' in function definition
```

### F632 - IsLiteral
```python
>>> f = 'foo'
>>> f is 'foo'
<stdin>:1: SyntaxWarning: "is" with a literal. Did you mean "=="?
True
```

### DIFF
--- a/docs/Coding-Conventions.md
+++ b/docs/Coding-Conventions.md
@@ -27,15 +27,20 @@ This is a living document that represents our coding conventions. Rules that are
 
 Additionally rules might be suffixed with one of the below:
 
-> 💻 - The convention is automatically enforced through tooling (I.e lint error)
+> 💻 - The convention is automatically enforced by `ni-python-styleguide` (By running `ni-python-styleguide lint ...`)
 >
-> ✨ - The convention is automatically fixed through tooling
+> ✨ - The convention is automatically fixed by `ni-python-stylgeuide` (`ni-python-styleguide` command doesn't exist yet)
 
 # This vs. other guides (like PEPs)
 
 This document serves as the single source of truth when it comes to Python coding conventions for NI code. Therefore other guides (such as the Google Python styleguide or various PEP-guides) are superseded by this one.
 
 In all cases where a convention comes from a PEP, it will be marked as such.
+
+# This vs. `ni-python-styleguide`
+
+Ideally, the conventions in this document would completely match the things `ni-python-styleguide` enforces.
+However, some checks we enforce don't correspond to conventions here as they either represent specific syntax issues or logic errors, both of which we assume the Python file is free from for the purposes of this document.
 
 ## Guides considered
 
@@ -114,7 +119,7 @@ line_with_99_chars = "spaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
 
 ```python
 # Bad - no spaces to form logical breaks in code
-def visit_argument_room(duration):
+def visit_argument_room(self, duration):
     start = datetime.now()
     while (datetime.now() - start) < duration:
         reply = self.argue_about_answer()
@@ -129,7 +134,7 @@ def visit_argument_room(duration):
 
 ```python
 # Good - use blank lines to separate code into logically-related sections
-def visit_argument_room(duration):
+def visit_argument_room(self, duration):
     start = datetime.now()
 
     while (datetime.now() - start) < duration:
@@ -148,7 +153,7 @@ def visit_argument_room(duration):
 
 ```python
 # Best - extract logic into well-named methods
-def visit_argument_room(duration):
+def visit_argument_room(self, duration):
     exit_reason = self._argue_about("answer", duration=duration)
 
     if exit_reason == "timeout":
@@ -225,20 +230,7 @@ class Outputs:
 
 > 🐍 This rule stems from [PEP 8](https://www.python.org/dev/peps/pep-0008)
 
-```python
-# Bad
-temporary_file
-```
-
-```python
-# Good
-temp_file
-```
-
-```python
-# Best
-tempfile
-```
+E.g. `tempfile` is preferred over `temp_file` or `temporary_file`
 
 ### [N.2.2] ✔️ **DO** Use `snake_case` for function, variable, and parameter names
 
@@ -489,22 +481,24 @@ except ImportError:
 
 ```python
 # Bad
-try:
-    # Too broad!
-    return eat(spam[key])
-except KeyError:
-    # Will also catch KeyError raised by eat()
-    return key_not_found(key)
+def lunchtime():
+    try:
+        # Too broad!
+        return eat(spam[key])
+    except KeyError:
+        # Will also catch KeyError raised by eat()
+        return key_not_found(key)
 ```
 
 ```python
 # Good
-try:
-    value = spam[key]
-except KeyError:
-    return key_not_found(key)
-else:
-    return eat(value)
+def lunchtime():
+    try:
+        value = spam[key]
+    except KeyError:
+        return key_not_found(key)
+    else:
+        return eat(value)
 ```
 
 ### [L.3.5] ❌ **DO NOT** Use flow control statements (`return`/`break`/`continue`) within the `finally` suite of a `try...finally`, where control flow would jump outside the `finally` suite
@@ -615,6 +609,23 @@ This includes setting `__all__` to the empty list if your module has no public A
 ```python
 # Good
 __all__ = ["spam", "ham", "eggs"]
+
+
+def spam():
+    pass
+
+
+def ham():
+    pass
+
+
+def eggs():
+    pass
+```
+
+```python
+# Good
+__all__ = []
 ```
 
 ### [L.7.2] ✔️ **DO** Prefix internal interfaces with a single leading underscore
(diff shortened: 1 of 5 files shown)


mshafer-NI:
To the question of calling out these rules in the docs, here's some thoughts:

F406 - ImportStarNotPermitted
 - maybe, it's against the style rules to do a wildcard import anyways... so probably fine to not explicitly doc, but also couldn't hur


F721 - DoctestSyntaxError
 - yay! this is the biggest concern against doctests (maybe add a call out in docs that doctests have valid syntax?? seems like this should be obvious...)

docs/Coding-Conventions.md
```diff
@@ -27,16 +27,21 @@ This is a living document that represents our coding conventions. Rules that are
...
 # This vs. other guides (like PEPs)
 
 This document serves as the single source of truth when it comes to Python coding conventions for NI code. Therefore other guides (such as the Google Python styleguide or various PEP-guides) are superseded by this one.
 
 In all cases where a convention comes from a PEP, it will be marked as such.
 
+# This vs. `ni-python-styleguide`
+
+Ideally, the conventions in this document would completely match the things `ni-python-styleguide` enforces.
+However, some checks we enforce don't correspond to conventions here as they either represent specific syntax issues or logic errors, both of which we assume the Python file is free from for the purposes of this document.
```
mshafer-NI, line 43:
```suggestion
However, some checks we enforce don't correspond to conventions here as they either represent specific syntax issues or logic errors. Both of which we assume the Python file is free from for the purposes of this document.
```

docs/Coding-Conventions.md
```diff
@@ -225,20 +230,7 @@ class Outputs:
...
-```python
-# Good
-temp_file
-```
-
-```python
-# Best
-tempfile
-```
+E.g. `tempfile` is preferred over `temp_file` or `temporary_file`
```
mshafer-NI, line 233: is this change intentional?

tests/test_convention_doc/test_codeblocks.py
```diff
@@ -23,7 +38,13 @@ def test_rule_codeblocks_documents_bad_good_best(codeblock):
 
 def test_bad_codeblocks_document_lint_errors(lint_codeblock, bad_codeblock):
     if bad_codeblock.rule.is_automatically_enforced:
-        result = lint_codeblock(bad_codeblock, "--format='%(code)s'")
+        result = lint_codeblock(
+            bad_codeblock,
+            "--format='%(code)s'",
+            ignore_unused_imports=(
+                "Import definitions that are not used" not in bad_codeblock.rule.header_text
```
mshafer-NI, line 45: hmmmm... that's definitely a magic string that could break this test if we ever tweak the header... could we key off something in the rule or in the expected error code??

mshafer-NI (changes requested):

DavidCurtiss (approved):

docs/Coding-Conventions.md
rtzoeller: How about this instead?

```suggestion
However, some checks we enforce don't correspond to conventions here as they either represent specific syntax issues or logic errors. We assume the Python file is free from both for the purposes of this document.
```

rtzoeller (approved):

docs/Coding-Conventions.md
thejcannon: Yes, the fact that they were codeblocks wasn't adding anything.

tests/test_convention_doc/test_codeblocks.py
thejcannon: Ha! I feel so dumb. Switched it to look for `F401` 😄

mshafer-NI (approved):

rick-morrow (approved):

kcuzner (approved):

irwand (approved):

sbethur (approved):

jryckman (approved):

neilvana (approved):

tests/test_convention_doc/test_codeblocks.py
```diff
@@ -1,11 +1,26 @@
...
+        codeblock,
+        *styleguide_args,
+        # We ignore unused import warnings as most of the codeblocks which produce this warning
+        # are just demonstrations of bad/good imports. Adding usage of the imports would add
+        # code to trivial examples, detracting from the interesting lines.
+        ignore_unused_imports=True,
+    ):
+
+        extend_ignore = [
+            # Undefined named. Defining all the names in each example would detract from the
```
pakdev, line 16: Undefined name. (remove d)

docs/Coding-Conventions.md
```diff
@@ -225,20 +230,7 @@ class Outputs:
```
pakdev, line 230: Apparently GitHub doesn't allow you to comment on lines a short distance away from changed lines, but I think "underscores are permissible, but should be avoided" is a bit harsh. The guidance on the pep-0008 page states "Function[/variable] names should be lowercase, with _words separated by underscores as necessary to improve readability._" That doesn't sound like "avoided" to me, rather, used judiciously.

pakdev, line 230: Apparently GitHub doesn't allow you to comment on lines more than a short distance away from changed lines, but I think "underscores are permissible, but should be avoided" is a bit harsh. The guidance on the pep-0008 page states "Function[/variable] names should be lowercase, with _words separated by underscores as necessary to improve readability._" That doesn't sound like "avoided" to me, rather, used judiciously.

pakdev (approved):

docs/Coding-Conventions.md
kcuzner: Isn't this just referring to package/module names?

alexweav (approved):

docs/Coding-Conventions.md
thejcannon: Yeah this convention is especially about naming packages/modules.
It comes from https://www.python.org/dev/peps/pep-0008/#package-and-module-names
> Modules should have short, all-lowercase names. Underscores can be used in the module name if it improves readability. Python packages should also have short, all-lowercase names, although the use of underscores is discouraged.

pakdev: Doh; the example looked like a variable - nm

thejcannon: Yeah, it was wrong of me to use code blocks previously 😄

alejandro5042 (approved):

stick152 (approved):

nethrasuresh (approved):

PR #33, opened 2020-11-19
Add flake8-docstrings linting

thejcannon:
This PR adds flake-docstrings to the flake8 plugins we use for enforcement.

NOTE: This PR does not choose a docstring format to enforce/use. We can do that in a follow-up PR. Since all of the error codes are enforced, this means there's some basic docstring stuff being enforced as well as more specific "if you're going to follow X, do it right" checks.

I reordered docstring conventions to roughly follow general conventions followed by more technical conventions
Note some checks are still "on" in the tool but not called out in the convention as they're kind of "syntactic" checks around the docstrings

NOTE: Our existing tooling ignores D1XX error codes. The convention is explicitly advocating for us to not ignore those codes as we should be documenting our code.

### DIFF
--- a/docs/Coding-Conventions.md
+++ b/docs/Coding-Conventions.md
@@ -887,8 +887,6 @@ This includes (if applicable) the function's:
 
 > 🐍 This rule stems from [PEP 257](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings)
 
-ℹ️ See the [Google docstring conventions for classes](https://google.github.io/styleguide/pyguide.html#384-classes)
-
 This includes (if applicable) the class's:
 
 - overall behavior
--- a/docs/Coding-Conventions.md
+++ b/docs/Coding-Conventions.md
@@ -853,19 +853,144 @@ def go_shopping():
 
 ## [D.1] Docstrings
 
-### [D.1.1] ✔️ **DO** Write docstrings for all public packages, modules, functions, classes, and methods
+### [D.1.1] ✔️ **DO** Write docstrings for all public packages, modules, functions, classes, and methods 💻
 
 > 🐍 This rule stems from [PEP 8](https://www.python.org/dev/peps/pep-0008) and [PEP 257](https://www.python.org/dev/peps/pep-0257/#what-is-a-docstring)
 
+> 💻 This rule is enforced by error codes D100-D107
+
 ℹ️ You can document a package by documenting the module docstring of the package directory's `__init__.py`
 
-### [D.1.2] ✔️ **DO** Put closing `"""` on the same line for one line docstrings
+### [D.1.2] ✔️ **DO** List exported modules and subpackages in a package's docstring
 
-> 🐍 This rule stems from [PEP 8](https://www.python.org/dev/peps/pep-0008) and [PEP 257](https://www.python.org/dev/peps/pep-0257/#one-line-docstrings)
+> 🐍 This rule stems from [PEP 257](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings)
 
-### [D.1.3] ✔️ **DO** Put closing `"""` on its own line for multiline docstrings
+### [D.1.3] ✔️ **DO** List relevant exported objects (classes, functions, exceptions, etc...) in a module's docstring
+
+> 🐍 This rule stems from [PEP 257](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings)
+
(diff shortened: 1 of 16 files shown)


jryckman (approved):

DavidCurtiss (approved):

irwand (approved):

tests/test_convention_doc/test_sections.py
```diff
@@ -17,5 +19,6 @@ def test_section_identifiers_unique(sections):
 
 
 def test_section_identifier_follows_case_convention(section):
+    """Test that the section's header starts with an uppercase latter."""
```
rtzoeller, line 22:
```suggestion
    """Test that the section's header starts with an uppercase letter."""
```

rtzoeller (approved):

sbethur (approved):

mshafer-NI (approved):

tests/test_convention_doc/doctypes.py
```diff
@@ -19,28 +24,70 @@ def from_text(cls, text, *, parent=None, header_depth=1):
...
+        """
         super().__init__(text, **kwargs)
         self.rules = Rule.from_text(text, parent=self, header_depth=3)
 
 
-class Rule(_RuleHeader):
+class Rule(_Region):
+    """A rule in the convention document.
+
+    Rules are the specific conventions within a subsection and are .
```
alexweav, line 80:
```suggestion
    Rules are the specific conventions within a subsection and are identified in the text by heading level 3.
```

alexweav (approved):

tests/test_convention_doc/doctypes.py
alexweav: This suggestion might be overly simplistic, since a rule includes more text than the h3 stuff alone.

thejcannon: oh wow, what an embarrassing catch. So you're edit is right, as they are _identified_ by H3 text. It's slightly semantic, but 🤷‍♂️ .

docs/Coding-Conventions.md
```diff
@@ -973,58 +973,101 @@ def sell(type_):
     """
 ```
 
-### [D.1.9] ✔️ **DO** List exported modules and subpackages in a package's docstring
+```python
+# Good
+class CheeseShop(object):
```
pakdev, line 978: Don't need `(object)`

kcuzner (approved):

neilvana (approved):

irwand (approved):

tests/test_convention_doc/doctypes.py
```diff
@@ -1,13 +1,18 @@
...
 import pathlib
 import re
 
 
-class _RuleHeader(object):
+class _Region(object):
     @classmethod
     def from_text(cls, text, *, parent=None, header_depth=1):
+        """Return the region from the given convention text.
+
```
pakdev, line 12: What about documentating the "text" parameter? Also, I thought I saw a discussion elsewhere about documentation styles. Is there a particular style we've adopted yet?

tests/test_convention_doc/test_rules.py
```diff
@@ -16,20 +16,24 @@
     ],
 )
 def enumerated_rule(request):
+    """Parametrized fixture of each rule along with its index in the parent subsection."""
```
pakdev, line 19: [D.1.9] ✔️ DO Write your docstrings as a command

pakdev (approved):

rick-morrow (approved):

tests/test_convention_doc/test_rules.py
thejcannon: Yeah `pytest` fixtures are kind of the rule breaker in this instance. I'll have to see how other people and tools handle them with regards to docstrings. And see how the group feels as well.

They essentially act like properties, which the docstring should be written as if you are documenting an attribute. Which is to say, you're describing the _thing_ not the _process_.

In this case, and until we decide otherwise I'm going to go with correctness over convention.

tests/test_convention_doc/doctypes.py
thejcannon: My plan is to follow-up this PR with one to take a stance on which convention style we use (the winds are blowing towards `sphynx`-style).

alejandro5042 (approved):

thejcannon:
FYI @stick152 and @nethrasuresh, submitting!

PR #34, opened 2020-11-19
Post tagging release workflow

rick-morrow:
This is an alternate workflow for: https://github.com/ni/python-styleguide/pull/31

In this workflow, the package version in the `pyproject.toml` file will always be set to `[version]-alpha`. 
The version in source is the next version that will be released.

When we are ready to release, we will use the [GitHub Release](https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/managing-releases-in-a-repository) UI to created a tagged release of the version currently in source, with a `v` prepended (which is just a GitHub standard).

So if we have `0.1.0-alpha` in source, we create a release for `v0.1.0`.

We probably want to create a draft release and have some manual step of reviewing everything to make sure the tag version matches the source version, but it's not necessary for the PyPI versioned package since that only looks at the `pyproject.toml` version.

### The Release workflow will do the following main steps:
1. Checkout the branch that the release was created for (by default, the checkout step would try to checkout the tagged branch created, which we don't want.).
1. Setup Python, setup Poetry Cache, install Poetry.
1. Use `poetry version patch` to update the version from `[version]-alpha` to `[version]` (i.e. `0.1.0-alpha` -> `0.1.0`).
1. Publish the package to PyPI.
1. Use `poetry version prepatch` to update the version from `[version]` to `[version+minorIncrement]-alpha` (i.e. `0.1.0` -> `0.1.1-alpha`).
1. Commit and submit the version bump to the branch the release was created for.

### DIFF
--- a/.github/workflows/Publish-Package.yml
+++ b/.github/workflows/Publish-Package.yml
@@ -30,6 +30,27 @@ jobs:
         with:
           poetry-version: 1.0.0
 
+      # @TODO: This is a workaround for there not being a way to check the lock file
+      #   See: https://github.com/python-poetry/poetry/issues/453
+      #   NOTE: This isn't ideal as it using the system Python and dirtying the system
+      #     Python libraries.
+      - name: Check for lock changes
+        run: |
+          PYTHONPATH="${PYTHONPATH}:${HOME}/.poetry/lib" \
+          python -c "from poetry.factory import Factory; \
+            locker = Factory().create_poetry('.').locker; \
+            exit(0) if locker.is_locked() and locker.is_fresh() else exit(1)" \
+            && echo 'OK'
+
+      - name: Install the Package
+        run: poetry install
+
+      - name: Lint the Code
+        run: poetry run ni-python-styleguide lint
+
+      - name: Run tests
+        run: poetry run pytest -v
+
       # If the version is 0.1.0-alpha.0, this will set the version to 0.1.0
       - name: Promote package version to release
         run: |
--- a/.github/workflows/Publish-Package.yml
+++ b/.github/workflows/Publish-Package.yml
@@ -26,11 +26,9 @@ jobs:
           key: ${{ runner.os }}-poetry-${{ hashFiles('**/poetry.lock') }}
           restore-keys: ${{ runner.os }}-poetry-
 
-      - name: Install Poetry
-        run: |
-          curl -fsS -o get-poetry.py https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py
-          python get-poetry.py -y
-          echo "$HOME/.poetry/bin" >> $GITHUB_PATH
+      - uses: Gr1N/setup-poetry@v4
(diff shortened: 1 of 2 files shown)


thejcannon:
Merge the latest master to fix the build

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "ni-python-styleguide"
-version = "0.1.0"
+version = "0.1.0-alpha.0"
```
thejcannon, line 3: We should probably add a comment that the alpha version here only exists to denote a "source"-based installation and the official releases are just Major.Minor.Update

.github/workflows/Publish-Package.yml
```diff
@@ -0,0 +1,62 @@
...
+          python-version: 3.6
+
+      - name: Set up Poetry cache for Python dependencies
+        uses: actions/cache@v2
+        with:
+          path: ~/.cache/pypoetry
+          key: ${{ runner.os }}-poetry-${{ hashFiles('**/poetry.lock') }}
+          restore-keys: ${{ runner.os }}-poetry-
+
+      - name: Install Poetry
```
thejcannon, line 29: Why not use the setup-poetry action?

thejcannon (commented):
Overall I like this "one-button release" approach. It does mean there's a brief window between tagging the release and having the next version be applied by the workflow. However this project doesn't see active changes quick enough for that to matter.

I want to point out that the "release" will have the alpha version in the source. On one hand, that's unfortunate. On the other I like that we'll be able to see from the version if an installation is from PyPI vs. from source (be it directly from GitHub or from a zipped release).

I think I like this. 👍

thejcannon:
One more thing... Should we be validating the release by running the PR pipeline (or running similar steps?)

thejcannon (commented):
One more thing... Should we be validating the release by running the PR pipeline (or running similar steps?)

.github/workflows/Publish-Package.yml
rick-morrow: Fixed

pyproject.toml
rick-morrow: I added some comments to clarify this

rick-morrow:
> 
> 
> One more thing... Should we be validating the release by running the PR pipeline (or running similar steps?)

Probably. I can look into splitting things into templates, or just pulling in the testing PR steps. 

Ideally we'd have a system where we didn't need to rerun tests on the same commit/package. I think the tests are fast enough that it shouldn't matter for now.

rick-morrow:
I took the easy way out and just re-added the testing steps: https://github.com/rick-morrow/python-styleguide/runs/1427138508?check_suite_focus=true

thejcannon (approved):
I'm excited to try this out when we're ready!

rtzoeller (approved):

mshafer-NI (approved):

PR #37, opened 2020-12-08
Auto acknowledge existing linter errors

mshafer-NI:
add a "acknowledge-existing-errors" that runs a lint and add a "noqa {code}" for most errors (blk101 ignored - just run black), to offending lines.

Objective is to make adoption or upgrading to newer version of the tool to be simple - "run lint, if many errors, fix what is reasonable and auto-supress remainder. This way new you can enforce any new code meets current standards, with minimal effort"


ni_python_styleguide/_cli.py
```diff
@@ -91,17 +93,7 @@ def main(ctx, verbose, quiet, config, exclude, extend_exclude):
...
-@click.option("--format", type=str, help="Format errors according to the chosen formatter.")
-@click.option(
-    "--extend-ignore",
-    type=str,
-    help="Comma-separated list of errors and warnings to ignore (or skip)",
-)
-@click.argument("file_or_dir", nargs=-1)
-@click.pass_obj
-def lint(obj, format, extend_ignore, file_or_dir):
+def _lint(obj, format, extend_ignore, file_or_dir):
```
mshafer-NI, line 96: moved body of lint to _lint and called from both - this was necessary to get the obj to pass correctly

tests/test_cli/test_acknowledge_existing_errors.py
```diff
@@ -0,0 +1,45 @@
...
+    'orange="""',
+    "lorem ipsum",
+    "dolor sit amet",
+    '"""',
+    "move_files()",
+]
+
+
+@pytest.mark.parametrize(
+    "input_lines,lineno,expected_in_mutiline",
```
rtzoeller, line 20: `multiline` is misspelled in a few places (`mutiline`, `mutliline`).

ni_python_styleguide/acknowledge_existing_errors/__init__.py
```diff
@@ -0,0 +1,70 @@
...
+
+    with open(filepath, mode="w") as out_file:
+        for line in code:
+            out_file.write(line)
+
+
+def acknowledge_lint_errors(lint_errors):
+    parsed_eorrors = [ni_python_styleguide.lint_errors_parser.parse(error) for error in lint_errors]
+
+    lint_errors_to_process = _filter_to_handled_errors(parsed_eorrors)
```
rtzoeller, line 42:
```suggestion
    parsed_errors = [ni_python_styleguide.lint_errors_parser.parse(error) for error in lint_errors]

    lint_errors_to_process = _filter_to_handled_errors(parsed_errors)
```

ni_python_styleguide/acknowledge_existing_errors/__init__.py
```diff
@@ -0,0 +1,70 @@
...
+    for error in lint_errors_to_process:
+        skip = 0
+        if _in_multiline_string(error_file=error.file, lineno=error.line):
+            # find when the multiline ends
+            skip = 1
+            while _in_multiline_string(error_file=error.file, lineno=error.line + skip):
+                skip += 1
+
+        cached_key = f"{error.file}:{error.line + skip}"
+        if error.code in handled_lines.get(cached_key, []):
```
thejcannon, line 54: It shouldn't be too difficult to write it to handle multiple codes, right? They just need to be comma-separated.

ni_python_styleguide/acknowledge_existing_errors/__init__.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+def _add_noqa_to_line(lineno, filepath, error_code, explanation):
+    with open(filepath, mode="r") as cache_file:
+        code = cache_file.readlines()
+
+    line = code[lineno]
+    line = line.rstrip("\n")
+    line += (
+        f"  # NOQA {error_code}: {explanation} - This suppression was "
```
thejcannon, line 29: I'd also prefer if we used lowercase `noqa` as that's what `flake8` uses in the documentation: https://flake8.pycqa.org/en/3.1.1/user/ignoring-errors.html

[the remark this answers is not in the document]
thejcannon: Or even the explanation... they might become stale.

People should be able to look up the error code descriptions.

tests/test_cli/test_acknowledge_existing_errors.py
```diff
@@ -0,0 +1,45 @@
...
+        (example_file_lines, 4, False),
+        (example_file_lines, 5, False),
+        (example_file_lines, 6, True),
+        (example_file_lines, 7, True),
+        (example_file_lines, 8, True),
+        (example_file_lines, 9, False),
+        (example_file_lines, 10, False),
+    ),
+)
+def test_can_acurately_detect_if_in_multiline_string(
```
thejcannon, line 34: This feels like an odd way of doing this. Seems like it can just be one test you run the function on each line for and assert the results are `[False, True, True, ...]`

ni_python_styleguide/acknowledge_existing_errors/__init__.py
```diff
@@ -0,0 +1,70 @@
...
+        "auto-generated to allow focus on handling new errors\n"
+    )
+    code[lineno] = line
+
+    with open(filepath, mode="w") as out_file:
+        for line in code:
+            out_file.write(line)
+
+
+def acknowledge_lint_errors(lint_errors):
```
thejcannon, line 39: I'm gonna need to see a bunch of tests for this function 😛 
Feel free to either declare source in multiline strings or in their own files in a subdir.

Some scenarios ideas:
- Multiline strings
- function with all params on one line
- function with params on an intermediate line
- function with params on different lines

weird stuff like that.

Could use snapshots to make it easier as well 🤔 In fact that'd be a good idea, you could have the tests that the inputs produce the outputs, but also have another set of tests that all of the outputs don't produce warnings when linted (therefore proving the noqas work).

ni_python_styleguide/_cli.py
```diff
@@ -117,3 +109,55 @@ def lint(obj, format, extend_ignore, file_or_dir):
...
+)
+@click.argument("file_or_dir", nargs=-1)
+@click.pass_obj
+def lint(obj, format, extend_ignore, file_or_dir):
+    _lint(obj=obj, format=format, extend_ignore=extend_ignore, file_or_dir=file_or_dir)
+
+
+class _Capturing(list):
+    """
+    context manager that captures stdout and presents as a list of the output
```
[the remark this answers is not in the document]
thejcannon: Other stuff is missing docstrings too, I'll let you run `flake8-docstrings ` on them instead of commenting 😄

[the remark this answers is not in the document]
thejcannon: Or pull latest `master` branch 🎉

ni_python_styleguide/acknowledge_existing_errors/__init__.py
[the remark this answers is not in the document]
mshafer-NI: for the comment - was going for the uber ugly-should-be-obvious-if-someone-tries-to-sneek-into-new-code
for the explanation, if it get's stale, the lines changed and should have been fixed.

ni_python_styleguide/_cli.py
```diff
@@ -117,3 +109,55 @@ def lint(obj, format, extend_ignore, file_or_dir):
...
+
+    from: https://stackoverflow.com/a/16571630
```
thejcannon, line 132: The stack overflow mentions https://docs.python.org/3/library/contextlib.html#contextlib.redirect_stdout

Would that work easier?

mshafer-NI: missed that edit, and yes, I'll switch to stdlib

ni_python_styleguide/acknowledge_existing_errors/__init__.py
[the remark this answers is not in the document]
thejcannon: It's a bold stance, but at this point I'd agree with you. Marking Resolved.

tests/test_cli/test_acknowledge_existing_errors.py
mshafer-NI: except that wouldn't be testing one thing (clean code principle), but most if not all conditions in a single assert.

mshafer-NI:
working on tests now

ni_python_styleguide/acknowledge_existing_errors/__init__.py
mshafer-NI: > function with all params on one line
> function with params on an intermediate line
> function with params on different lines

these don't trigger a non-BLK100 code, do you have examples in mind?

thejcannon: Dang we might have to hold off until we get `pep8-naming` incorporated 😢 

You could short-circuit that by adding it to our dev dependencies and then ignoring all it's errors when we lint ourselves, and then turn it back on for this test.

Or you can add a local flake8 plugin and use it here.

Up to you.

ni_python_styleguide/acknowledge_existing_errors/__init__.py
mshafer-NI: clarified the warning that this was for the same code on the same line, and added explicitly handling multiple codes via a comma seperated list (see expected_output files)

pyproject.toml
```diff
@@ -31,16 +31,41 @@ flake8-docstrings = "^1.5.0"
...
 pytest_click = "^1.0.2"
+black = "^20.8b1"
+pytest-snapshot = "^0.4.2"
 
 [tool.poetry.scripts]
 ni-python-styleguide = 'ni_python_styleguide._cli:main'
 
 [tool.black]
 line-length = 100
+exclude = '''
```
thejcannon, line 42: We should only list the directories we expect to exclude (E.g. not `.hg` or `.tox`, etc...) and maybe also not have a `foo.py` in there 😄

tests/test_cli/test_acknowledge_existing_errors.py
```diff
@@ -0,0 +1,118 @@
...
+
+    def runner(base_args=[], lint_args=[]):
+        return styleguide(*base_args, "acknowledge-existing-violations", *lint_args)
+
+    chdir(str(tmp_path))
+
+    return runner
+
+
+@pytest.fixture
```
thejcannon, line 78: This should be moved up to `conftest.py` and shared with the other tests (so it isn't duplicated).

Perhaps instead of a fixture per command, we instead provide a command-runner-factory fixture? Developer's choice.

tests/test_cli/test_acknowledge_existing_errors.py
```diff
@@ -0,0 +1,118 @@
...
+    and passed on the cmd line in the following order: `<cmd> <base_args> lint <lint_args>.
+    """
+
+    def runner(base_args=[], lint_args=[]):
+        return styleguide(*base_args, "lint", *lint_args)
+
+    return runner
+
+
+@pytest.mark.parametrize("test_dir", [x for x in TEST_CASE_DIR.iterdir() if x.is_dir()])
```
thejcannon, line 92: Instead of subdirectories for each test, can you we use a file for the input and for the output?

E.g. the snapshots dir contains:

```
tescase1_input.py
testcase1_output.py
testcase2_input.py
...
```

(with a better name than "testcase")

tests/test_cli/test_acknowledge_existing_errors.py
```diff
@@ -0,0 +1,118 @@
...
+def test_given_bad_input_produces_expected_output(
```
thejcannon, line 93: The current strategy makes it appear as though we'll be making new test cases for each possible error, which is something I don't think we should be doing (or want other developers to think they have to do when adding new rules/plugins).

Since the tool operates on lines, I would recommend just testing scenarios that cover a range of possible _line locations_.

E.g. A test for a single-line statement (could be an import, could be an assignment, whatever you want). A test for an error in a multiline comment. A test for a parameter name with (a single line function, a triple-line function, and an N-line function). etc...

tests/test_cli/test_acknowledge_existing_errors.py
mshafer-NI: the folder per test yields much easier path manipulations (list dir, input is "", expected output is "" vs, list dir, filter to files with `*_input`, assert matches `$1_output`) and provides ordering/speration regardless of what the explorer list order is set to (file based is dependent on by-name for input/expected to pair up)

tests/test_cli/test_acknowledge_existing_errors.py
mshafer-NI: was just trying to come up with lots of examples/tests with some organization..., 

could re-arrange into something like.
* import_line_tests
* method_doc_line_tests
* method_signature_tests
* in_method_tests

thoughts?

thejcannon: Seems reasonable. But might as well simplify and s/method/function.

Something to add to the import tests pile: single-line import vs. triple-line import vs. N-line import (akin to the function equivalent).

neilvana (approved):

neilvana:
Please resolve the issues brought up by Josh, but I'm good with the changes.

tests/test_cli/test_acknowledge_existing_errors.py
mshafer-NI: made a generic "styleguide_command" and update `test_lint` to use it.

ni_python_styleguide/acknowledge_existing_errors/__init__.py
mshafer-NI: lots of tests added, but I'm going to hold off on submitting this PR until we have pep8-naming in, you can see my testing with that turned on [in my dev branch](https://github.com/mshafer-NI/ni-python-styleguide/blob/auto_noqa_re_arrange_tests/tests/test_cli/acknowledge_existing_errors_test_cases__snapshots/function_signature_tests/expected_output.py).

ni_python_styleguide/__init__.py
```diff
@@ -0,0 +1 @@
+"""NI's internal and external style rules enforcement tool for Python."""
```
mshafer-NI, line 1: running doctests via pytest made import errors unless this file was added

pyproject.toml
```diff
@@ -31,16 +31,32 @@ flake8-docstrings = "^1.5.0"
...
+    | \.git          # root of the project
+    | \.venv
+  )/
+  | /.*__snapshots/.*_output\.py # exclude the snapshot outputs
+)
+'''
 
 [tool.pytest.ini_options]
-addopts = "--doctest-modules --ignore-glob='ni_python_styleguide/**'"
+addopts = "--doctest-modules"
```
mshafer-NI, line 54: actually run the doctests in our module

ni_python_styleguide/acknowledge_existing_errors/__init__.py
mshafer-NI: updated - see snapshots for examples

ni_python_styleguide/__init__.py
thejcannon: Apparently `pytest` still requires `__init__.py` files :/

thejcannon (commented):
I'll try and look at these Monday 😄 

Thanks for your patience

ni_python_styleguide/__init__.py
mshafer-NI: yep :/

ni_python_styleguide/_acknowledge_existing_errors/__init__.py
```diff
@@ -0,0 +1,88 @@
...
+    line = line.rstrip("\n")
+
+    existing_suppression = re.search(r"noqa (?P<existing_suppresions>[\w\d]+\: [\w\W]+?) -", line)
+    if existing_suppression:
+        before = existing_suppression.groupdict()["existing_suppresions"]
+        if error_code not in before:
+            line = line.replace(before, before + f", {error_code}: {explanation}") + "\n"
+    else:
+        line += (
+            f"  # noqa {error_code}: {explanation} - This suppression was "
```
thejcannon, line 38: How would you feel about shortening this to `# noqa X100: Foo bar baz (auto-generated noqa)`?

Still has the same elements without being _overly_ verbose about it.

ni_python_styleguide/_acknowledge_existing_errors/__init__.py
```diff
@@ -0,0 +1,88 @@
...
+import ni_python_styleguide._lint_errors_parser
+
+
+def _filter_to_handled_errors(lint_errors):
+    not_handled_errors = {"BLK100"}
+    return filter(lambda o: o.code not in not_handled_errors, lint_errors)
+
+
+def _in_multiline_string(error_file, lineno):
+    with open(error_file) as in_file:
```
thejcannon, line 14: You're opening and parsing the file multiple times, might as well just open/parse once and keep the contents in memory. Have that be passed into this function.

Feel free to `del` the variable to free up the memory if that's a concern.

ni_python_styleguide/_acknowledge_existing_errors/__init__.py
```diff
@@ -0,0 +1,88 @@
...
+            out_file.write(line)
+
+
+def acknowledge_lint_errors(lint_errors):
+    """Add a "noqa" comment for each of existing errors (unless excluded).
+
+    Excluded error (reason):
+    BLK100 - run black
+    """
+    parsed_errors = filter(
```
thejcannon, line 56: Consider using `map`, as well as the boolean-ized `filter`.

```python
parsed_errors = map(ni_python_styleguide._lint_errors_parser.parse, lint_errors)
parsed_errors = filter(None, parsed_errors)
lint_errors_to_process = [error for error in parsed_errors if error not in IGNORE_ERRORS]
```

ni_python_styleguide/_acknowledge_existing_errors/__init__.py
```diff
@@ -0,0 +1,88 @@
+from collections import defaultdict
+import logging
+import re
+
```
thejcannon, line 4: I've always preferred `pathlib` with it's read/write semantics over `with open() as foo: ...`

Makes for easier to read code as there's less branching:
`pathlib.Path(...).read_text()` 😄

ni_python_styleguide/_acknowledge_existing_errors/__init__.py
```diff
@@ -0,0 +1,88 @@
...
+    # if occurances of multiline string markers is odd, this must be in a multiline
+    return count_of_multiline_starts_and_stops % 2 == 1
+
+
+def _add_noqa_to_line(lineno, filepath, error_code, explanation):
+    with open(filepath, mode="r") as cache_file:
+        code = cache_file.readlines()
+
+    line = code[lineno]
+    line_had_newline = line.endswith("\n")
```
thejcannon, line 28: When would a line not have a newline? End of File?
---
Another way of doing this is to make some variable that holds onto "line ending" it'll be `""` or `"\n"` and then unconditionally add it at the end.

ni_python_styleguide/_acknowledge_existing_errors/__init__.py
```diff
@@ -0,0 +1,88 @@
...
+    else:
+        line += (
+            f"  # noqa {error_code}: {explanation} - This suppression was "
+            "auto-generated to allow focus on handling new errors\n"
+        )
+    if not line_had_newline:
+        line.rstrip("\n")
+    code[lineno] = line
+
+    with open(filepath, mode="w") as out_file:
```
thejcannon, line 45: Again, we're opening-and-writing in a loop 😢

ni_python_styleguide/_cli.py
```diff
@@ -1,10 +1,12 @@
...
-
+import contextlib
 import flake8.main.application
-
+import logging
+import pathlib
 import toml
+from io import StringIO
+
+from . import _acknowledge_existing_errors
```
thejcannon, line 9: `from ni_python_styleguide import ...`

ni_python_styleguide/_lint_errors_parser/__init__.py
```diff
@@ -0,0 +1,58 @@
+import re
```
thejcannon, line 1: This should probably just be a module if we aren't putting any other files in the directory (same for the other dir)

tests/test_cli/test_acknowledge_existing_errors.py
thejcannon: OK. How would you feel about renaming to `input.py` and `expected.py` for brevity, then?

thejcannon (approved):
I worry that we haven't isolated the snapshot test environment enough, and when we add new plugins we might add noise to the reviews by modifying the input/output. However, that's a problem we can solve later, and this is really good stuff and I'd like to see it get in.

irwand (approved):

PR #38, opened 2021-01-06
Bump the package version to 0.1.1-alpha

rick-morrow:
When testing out the Poetry publish, I accidentally pushed the 0.1.0 version to PyPI. Unfortunately, even if you delete a release on PyPI, it is still indexed and you cannot overwrite it.

To get around this, we'll just have to bump the version and push a new package.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,7 +3,7 @@ name = "ni-python-styleguide"
 # The -alpha.0 here denotes a source based version
 # This is removed when released through the Publish-Package.yml GitHub action
 # Official PyPI releases follow Major.Minor.Patch
-version = "0.1.0-alpha.0"
+version = "0.1.1-alpha.0"
 description = "NI's internal and external Python linter rules and plugins"
 authors = ["NI <opensource@ni.com>"]
 license = "MIT"


mshafer-NI (approved):

thejcannon:
For clarification: This should be automatically handled by the GitHub Action, however we'll never get to the step that does this, as it is after the publish step, which we expect to fail, right?

rick-morrow:
> 
> 
> For clarification: This should be automatically handled by the GitHub Action, however we'll never get to the step that does this, as it is after the publish step, which we expect to fail, right?

Yes, in the future we will just need to create the GitHub release and version incrementing will happen automatically, at least for minor version bumps.

For any other version bumps, we will need to manually update the version like this PR. (i.e 0.2.0, 1.0.0)

mshafer-NI:
> For clarification: This should be automatically handled by the GitHub Action, however we'll never get to the step that does this, as it is after the publish step, which we expect to fail, right?

yes - we expect normal for the workflow to publish the currently listed version (less alpha) and bump, but it is currently pointing at a burned number

thejcannon (approved):

DavidCurtiss:
> even if you delete a release on PyPI, it is still indexed and you cannot overwrite it.

However, I think you can delete/recreate the whole project and it'll wipe out the release. Since this was the first release, that would be a reasonable option.

PR #39, opened 2021-01-07
Use admin PAT to bypass checks when bumping Poetry version

rick-morrow:
The workflow to increment the Poetry version [isn't working](https://github.com/ni/python-styleguide/runs/1658003895?check_suite_focus=true) because the branch has protections requiring reviewers to approve commits before they're submitted (usually through PR).

Apparently, this is a [known issue](https://github.com/ad-m/github-push-action/issues/36).
This is by design https://github.community/t/how-to-push-to-protected-branches-in-a-github-action/16101/5

Some workarounds are:
1. (This PR) Create a PAT from an admin account that is exempted from these protections.
2. Create a workflow that bumps the Poetry version and creates a PR. Reviewers will need to review and push the changes through.

I'm opting for 1 since I think we would want this process automated.

Some disclaimers:
Exposing an admin PAT is somewhat of a vulnerability since any of our workflows can use it. I think it is okay in this case because to submit a workflow, it would need to go through the review process and should be something we can catch.

Also, I'm including the Poetry version bump to get us in line with the current PyPI release since this failed in the previous workflow

### DIFF
--- a/.github/workflows/Publish-Package.yml
+++ b/.github/workflows/Publish-Package.yml
@@ -49,7 +49,6 @@ jobs:
         run: poetry run ni-python-styleguide lint
 
       - name: Run tests
-
         run: poetry run pytest -v
 
       # If the version is 0.1.0-alpha.0, this will set the version to 0.1.0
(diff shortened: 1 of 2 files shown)


mshafer-NI (approved):
sounds good to me

thejcannon (approved):

PR #42, opened 2021-01-08
Recommend Google as the docstring format

thejcannon:
Having any documentation is better than no documentation. Likewise following a format is better than not. However choosing a format can be intractable.

We chose Google style because the majority of people preferred it.

Since type hints exist (and we will be recommending their usage) the type information format of either doesn't concern me much.

NOTE: This PR started off recommending sphinx format

### DIFF
--- a/docs/Coding-Conventions.md
+++ b/docs/Coding-Conventions.md
@@ -861,6 +861,13 @@ def go_shopping():
 
 ℹ️ You can document a package by documenting the module docstring of the package directory's `__init__.py`
 
+### Which docstring format should I follow?
+
+We recommend (and use internally) using the [Sphinx docstring format](https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html#the-sphinx-docstring-format)
+but you can choose any format so long as you are consistent.
+
+(We chose the Sphinx docstring format because we're using `spinx`, allowing us to use docstrings to autogenerate rich documentation)
+
 ### [D.1.2] ✔️ **DO** List exported modules and subpackages in a package's docstring
 
 > 🐍 This rule stems from [PEP 257](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings)
--- a/docs/Coding-Conventions.md
+++ b/docs/Coding-Conventions.md
@@ -866,7 +866,7 @@ def go_shopping():
 We recommend (and use internally) using the [Sphinx docstring format](https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html#the-sphinx-docstring-format)
 but you can choose any format so long as you are consistent.
 
-(We chose the Sphinx docstring format because we're using `spinx`, allowing us to use docstrings to autogenerate rich documentation)
+(We chose the Sphinx docstring format because we're using `sphinx`, allowing us to use docstrings to autogenerate rich documentation)
 
 ### [D.1.2] ✔️ **DO** List exported modules and subpackages in a package's docstring
 
--- a/docs/Coding-Conventions.md
+++ b/docs/Coding-Conventions.md
@@ -863,10 +863,7 @@ def go_shopping():
 
 ### Which docstring format should I follow?
 
-We recommend (and use internally) using the [Sphinx docstring format](https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html#the-sphinx-docstring-format)
-but you can choose any format so long as you are consistent.
-
-(We chose the Sphinx docstring format because we're using `sphinx`, allowing us to use docstrings to autogenerate rich documentation)
+We recommend (and use internally) using the [Google docstring format](https://google.github.io/styleguide/pyguide.html#383-functions-and-methods) but you can choose any format so long as you are consistent.
 
 ### [D.1.2] ✔️ **DO** List exported modules and subpackages in a package's docstring
 
--- a/ni_python_styleguide/_cli.py
+++ b/ni_python_styleguide/_cli.py
@@ -38,7 +38,7 @@ class AllowConfigGroup(click.Group):
     """click.Group subclass which allows for a config option to load options from."""
 
     def __init__(self, *args, **kwargs):
-        """Construct the click.Group with the config option."""
+        """Constructs the click.Group with the config option."""
         kwargs["params"].append(
             click.Option(
                 ["--config"],
--- a/ni_python_styleguide/config.ini
+++ b/ni_python_styleguide/config.ini
@@ -65,6 +65,13 @@ ignore =
     F812
 
     # flake8-docstrings
+    # Ignoring the requirement on imperative mood, as not all conventions agree on this
+    # and ultimately is up to the developer. If, in the future, the mood was enforced based on
+    # convention choice we can consider removing this supression as that would make documentation
+    # consistent within a standard (but maybe not a codebase)
+    # Developers are welcome to turn back on this error in their specific projects.
+    # The same goes for the "descriptive"-mood error, if one was to be added to pydocstyle.
+    D401
     # Ignoring things enforced by black
     D200  # One-line docstring should fit on one line with quotes
     D206  # Docstring should be indented with spaces, not tabs
@@ -83,4 +90,4 @@ ignore =
 extend-ignore=C90
 
 # flake8-docstrings
-docstring-convention=google
+docstring-convention=all
--- a/ni_python_styleguide/config.ini
+++ b/ni_python_styleguide/config.ini
@@ -83,4 +83,4 @@ ignore =
 extend-ignore=C90
 
 # flake8-docstrings
-docstring-convention=all
+docstring-convention=google
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -10,14 +10,14 @@ from ni_python_styleguide.__main__ import main as styleguide_main
 
 
 def pytest_collection_modifyitems(items):
-    """Ignore deprecation warnings in all tests."""
+    """Ignores deprecation warnings in all tests."""
     for item in items:
         item.add_marker(pytest.mark.filterwarnings("ignore::DeprecationWarning"))
 
 
 @pytest.fixture
 def styleguide(monkeypatch, cli_runner):
-    """Fixture which runs the styleguide when executed.
+    """Runs the styleguide when executed.
 
     Args passed to the function are equivalent to CLI args,
     and are automatically stringified.
@@ -40,7 +40,7 @@ def styleguide(monkeypatch, cli_runner):
 
 @pytest.fixture
 def chdir():
-    """Fixture which changes the current working directory when executed."""
+    """Changes the current working directory when executed."""
     cwd = os.getcwd()
     yield os.chdir
     os.chdir(cwd)
--- a/tests/test_cli/test_bare.py
+++ b/tests/test_cli/test_bare.py
@@ -2,7 +2,7 @@
 
 
 def test_no_args_prints_help(styleguide):
-    """Test that running the styleguide without any arguments prints the help info."""
+    """Tests that running the styleguide without any arguments prints the help info."""
     result = styleguide()
 
     assert result
@@ -10,7 +10,7 @@ def test_no_args_prints_help(styleguide):
 
 
 def test_help_prints_help(styleguide):
-    """Test that running the styleguide with "--help" prints the help info."""
+    """Tests that running the styleguide with "--help" prints the help info."""
     result = styleguide("--help")
 
     assert result
--- a/tests/test_cli/test_lint.py
+++ b/tests/test_cli/test_lint.py
@@ -11,7 +11,7 @@ TOO_LONG_LINE = "a_really_long_order = [" + ", ".join(itertools.repeat('"spam"',
 
 @pytest.fixture
 def styleguide_lint(styleguide, tmp_path, chdir):
-    """Fixture which will run the styleguide with the "lint" subcommand.
+    """Runs the styleguide with the "lint" subcommand.
 
     Both `base_args` and `lint_args` must be iterables which will be transformed into strings
     and passed on the cmd line in the following order: `<cmd> <base_args> lint <lint_args>.
@@ -29,7 +29,7 @@ def styleguide_lint(styleguide, tmp_path, chdir):
 
 @pytest.fixture
 def config_write(tmp_path):
-    """Fixture which will write the specified data to a config file for the tool."""
+    """Writes the specified data to a config file for the tool."""
 
     def writer(**config_data):
         (tmp_path / "pyproject.toml").write_text(
@@ -48,7 +48,7 @@ def config_write(tmp_path):
     ]
 )
 def styleguide_lint_with_options(request, styleguide_lint, config_write, tmp_path):
-    """Fixture which will run the styleguide with the "lint" subcommand, given the provided options.
+    """Runs the styleguide with the "lint" subcommand, given the provided options.
 
     The options will be tested on the command line as well as in the config.
     `base_args` and `lint_args` are similar to the `styleguide_lint` fixture, with the exception of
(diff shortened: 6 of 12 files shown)


docs/Coding-Conventions.md
```diff
@@ -861,6 +861,13 @@ def go_shopping():
 
 ℹ️ You can document a package by documenting the module docstring of the package directory's `__init__.py`
 
+### Which docstring format should I follow?
+
+We recommend (and use internally) using the [Sphinx docstring format](https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html#the-sphinx-docstring-format)
+but you can choose any format so long as you are consistent.
+
+(We chose the Sphinx docstring format because we're using `spinx`, allowing us to use docstrings to autogenerate rich documentation)
```
mshafer-NI, line 869:
```suggestion
(We chose the Sphinx docstring format because we're using `sphinx`, allowing us to use docstrings to autogenerate rich documentation)
```

mshafer-NI (approved):
I'm good with this rationale.

rtzoeller (approved):

irwand (approved):

DavidCurtiss:
You can use RST syntax in Google-style docstrings, too (e.g. [here](https://python-docs.systemlink.io/en/stable/_modules/systemlink/clients/tag/_tag_data_update.html#TagDataUpdate), generating [this](https://python-docs.systemlink.io/en/stable/api_reference/tag.html#systemlink.clients.tag.TagDataUpdate)).
And I don't mind the indentation -- it helps readability.

docs/Coding-Conventions.md
```diff
@@ -861,6 +861,13 @@ def go_shopping():
 
 ℹ️ You can document a package by documenting the module docstring of the package directory's `__init__.py`
 
+### Which docstring format should I follow?
+
+We recommend (and use internally) using the [Sphinx docstring format](https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html#the-sphinx-docstring-format)
```
neilvana, line 866: We have used Google format for much of our documentation internally and externally : see  https://github.com/ni/nimi-python/blob/8de853ace8b796145814e1b8119233ba4dff7df1/generated/nidmm/nidmm/session.py#L588 for example.  The Google plugin is built into Sphinx and it effectively converts the format the sphinx format before rendering the documentation so you can still use all the other features of Sphinx.  I would recommend we stick with Google format since it is significantly more readable in my opinion.

thejcannon: (I'll use this for the main discussion).

As @DavidCurtiss points out, since Sphinx is generating the documentation, it will always handle RST-directives 👍. So the choice is simply about style (I'll update the PR accordingly).

So let's use your linked code for comparison (adjusted to match our conventions):

Google format
```python
        """Registers a callback function to receive an event when the specified hardware event occurs.

        When you stop a task explicitly any pending events are discarded. For example, if you call 
        DAQmx Stop Task then you do not receive any pending events.

        Args:
            signal_type: Specifies the type of signal for which you want to receive results.
            callback_method: Specifies the function that you want DAQmx to call when the event 
                occurs. The function you pass in this parameter must have the following prototype:

                >>> def callback(task_handle, signal_type, callback_data):
                >>>     return 0

                Upon entry to the callback, the task_handle parameter contains the handle to the 
                task on which the event occurred. The signal_type parameter contains the integer 
                value you passed in the signal_type parameter of this function. The callback_data
                parameter contains the value you passed in the callback_data parameter of this 
                function.

                Passing None for this parameter unregisters the event callback function.
        """
```

Sphinx format
```python
        """Register a callback function to receive an event when the specified hardware event occurs.

        When you stop a task explicitly any pending events are discarded. For example, if you call 
        DAQmx Stop Task then you do not receive any pending events.

        :param signal_type: Specifies the type of signal for which you want to receive results.
        :param callback_method: Specifies the function that you want DAQmx to call when the event 
            occurs. The function you pass in this parameter must have the following prototype:

            >>> def callback(task_handle, signal_type, callback_data):
            >>>     return 0

            Upon entry to the callback, the task_handle parameter contains the handle to the task on 
            which the event occurred. The signal_type parameter contains the integer value you 
            passed in the signal_type parameter of this function. The callback_data parameter 
            contains the value you passed in the callback_data parameter of this function.

            Passing None for this parameter unregisters the event callback function.
        """
```

The difference being a tab's worth of spaces on a description's subsequent lines (due to Google's "Args").

There is this quote by Guido (on the subject of using `StructuredText` for docstrings) someone included in PEP 287 I think is somewhat relevant:

> I still think that using indentation to indicate sectioning is wrong. If you look at how real books and other print publications are laid out, you'll notice that indentation is used frequently, but mostly at the intra-section level. Indentation can be used to offset lists, tables, quotations, examples, and the like. (The argument that docstrings are different because they are input for a text formatter is wrong: the whole point is that they are also readable without processing.)
>
> I reject the argument that using indentation is Pythonic: text is not code, and different traditions and conventions hold. People have been presenting text for readability for over 30 centuries. Let's not innovate needlessly.

---

Since this is a matter of style, however I think we can settle this like line length and put it to a vote.

kcuzner (approved):

docs/Coding-Conventions.md
DavidCurtiss: "If you look at how real books and other print publications are laid out ..."
In those examples, they use enlarged text, etc. as section headers. We have no such ability here. Theoretically we could use Markdown to add section headers, to avoid indentation, but that is not among the options we've been presented with here.

My vote is Google style.

thejcannon: > My vote is Google style.

I'm going to send out a form 😛

thejcannon: I'll add to this discussion that the Sphinx format is likely to match our other language conventions to some degree.

I've seen our Javascript convention uses `@param` and `@returns` in a manner similar to Sphinx.
image

(Will edit with other languages)

alejandro5042: I voted for Google style in the form.

IMHO it is the most readable doc style _as displayed in a text editor_, which is how I read docs ~90% of the time (and I imagine I'm not alone).

From [Napoleon - Marching toward legible docstrings — napoleon 0.7 documentation](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/):

image

If there was a way to format with Markdown that was at least growing in popularity, I would lean toward that :)

thejcannon: @alejandro5042 Both examples listed provide types, which aren't necessary since we have type hints. The approproate comparison would be:

```
:param path: The path of the file to wrap
:param field_storage: The :class:`FileStorage` instance to wrap
:param temporary: Whether or not to delete the file when the File
   instance is destructed
:returns: A buffered writable file descriptor
```

vs.

```
Args:
    path: The path of the file to wrap
    field_storage: The :class:`FileStorage` instance to wrap
    temporary: Whether or not to delete the file when the File
       instance is destructed

Returns:
    A buffered writable file descriptor
```

The difference being the extra indentation and how you specify the section (with Sphinx just "jumping into it")

thejcannon: Well the results are in, and it appears people have a slight affinity for Google format (6 Google, 5 Sphinx, 2 ambivalent).

Unless someone changes their mind I'll switch the PR over and reset votes.

mshafer-NI (approved):

ni_python_styleguide/_cli.py
```diff
@@ -38,7 +38,7 @@ class AllowConfigGroup(click.Group):
     """click.Group subclass which allows for a config option to load options from."""
 
     def __init__(self, *args, **kwargs):
-        """Construct the click.Group with the config option."""
+        """Constructs the click.Group with the config option."""
```
DavidCurtiss, line 41: Hmm.. I like the _styling_ of the Google docstrings, but I've tended to prefer imperative method summaries, to harmonize with the rest of the Python community. But I can live with descriptive instead, if we have to go whole-hog on one standard or the other. One advantage of descriptive-style is that it would match our C# docstring conventions, so that's good, at least.

docs/Coding-Conventions.md
```diff
@@ -861,6 +861,10 @@ def go_shopping():
 
 ℹ️ You can document a package by documenting the module docstring of the package directory's `__init__.py`
 
+### Which docstring format should I follow?
+
+We recommend (and use internally) using the [Google docstring format](https://google.github.io/styleguide/pyguide.html#383-functions-and-methods) but you can choose any format so long as you are consistent.
```
rtzoeller, line 866:
```suggestion
We recommend (and internally use) the [Google docstring format](https://google.github.io/styleguide/pyguide.html#383-functions-and-methods) but you can choose any format so long as you are consistent.
```

IMO this reads better, but feel free to reject.

ni_python_styleguide/_cli.py
irwand: well shit, I prefer the other style.. Must we adopt this as part of Google docstring style?
I suppose if we subscribe to google docstring, we should just follow that and keep it simple.. but.. bah!

irwand (approved):

ni_python_styleguide/_cli.py
thejcannon: Yeah, it's gonna be all-or-nothing. 😄 Anything in the middle would just be confusing...

DavidCurtiss (approved):

ni_python_styleguide/_cli.py
sbethur: bummer :( ..anyway not a big deal.

sbethur (approved):

alexweav (approved):

ni_python_styleguide/_cli.py
pakdev: Well, it's kind of a big deal. Now our conventions violate PEP-0257. If this difference was highlighted pre-voting I would have changed from ambivalent to Sphinx.

irwand: is another voting called for ? I really don't like this change.. I can live with it.. but PEP > google docs standard IMO.. ?

thejcannon: Well, the original vote was close: 6-to-5. Are people feeling like another vote might be in order?
Anyone else want to chime in?
@neilvana @alejandro5042 @mshafer-NI @rtzoeller

thejcannon: (Names are just people who are vocal, not those who voted for Google)

mshafer-NI: Well, this doesn't change my vote (although I agree that enacted/approved PEP's should take precedent)

thejcannon: > If this difference was highlighted pre-voting I would have changed from ambivalent to Sphinx.

I suppose that would change the count from 6-5 to 6-6, which is a tie.

DavidCurtiss: It looks like everyone that's in favor of the Google style (or at least everyone that's spoken up) is also in favor of employing a PEP-257 exception to it.

(And if we revote, that should be included. Maybe with a ranked-choice vote?)

Or maybe our vote split just means that we should pick whichever we prefer, but just be consistent within the project.

alejandro5042: The example provided by the [Google conventions doc we link to](https://google.github.io/styleguide/pyguide.html#383-functions-and-methods) states:

> A docstring should give enough information to write a call to the function without reading the function’s code. The docstring should be descriptive-style ("""Fetches rows from a Bigtable.""") rather than imperative-style ("""Fetch rows from a Bigtable."""),

The difference is: _Fetches vs. Fetch_

I'd be happy with either, and neither IMHO negates using the Google style for formatting your docs.

I'm wondering: how important is this distinction? I guarantee you I will never think about this again after this PR 😛  Persuading devs to slow down and document their code is hard enough even before we become grammar police and enforce verb tense to use 😄 I'd rather tell devs "please document! and here's how it should look" than "please learn the difference between imperative and descriptive and keep that in mind when you document." I'd rather advocate active verb tense over passive.

alejandro5042 (approved):

alejandro5042 (approved):

ni_python_styleguide/_cli.py
pakdev: Is there a linter that enforces descriptive-style? I only see an error (D401) for not being imperative-style in [pydocstyle](http://www.pydocstyle.org/en/stable/error_codes.html) (though you'd think the reverse would be trivial to implement). The more we can enforce consistency the better.

thejcannon: > It looks like everyone that's in favor of the Google style (or at least everyone that's spoken up) is also in favor of employing a PEP-257 exception to it.

It's doable, but I'm not a fan of linking to external documentation and then listing exceptions. Ideally readers/clients don't need a mental map of rules and their locations, and the tools help enforce our policies.

(It's minor, but I'm also personally not a fan of mixing the imperative style description which can be seen as "first-party" with the "third-party" "Args", "Returns", "Raises", etc...)

> Or maybe our vote split just means that we should pick whichever we prefer, but just be consistent within the project.

This is technically possible (and is what we're doing pre-this PR by enforcing _all_ styles), however I don't think we should allow ourselves to bifurcate our styles.

> I'm wondering: how important is this distinction?

In the grand scheme, not important. But with tooling we'll want to make a choice to be consistent. 

> Is there a linter that enforces descriptive-style?

I sniffed around and didn't find one. I assume `pydocstyle` doesn't have one because it somewhat closely follows PEP-257 (although the fact that it has a "google" format option makes me think it could have one but just doesn't).

I went ahead and opened an issue: https://github.com/PyCQA/pydocstyle/issues/521

neilvana: I wish the Python community had a readable, more consistent, standard.  I like the overall readability of Google Style because it looks like it was written for a human and not a computer.  The Sphinx style seems written for a parser not a human.  I prefer Google Style for the same reason I prefer yaml or toml over XML.

But why did Google decide to deviate from PEP-257 in this way? --Neil wrings his hands-- Did they want to match other languages at Google?  Were they just trying to be difficult?  I don't think I even noticed this deviation when reading the style the first time (so good catch) and have generally preferred imperative mood since it usually is shorter and matches convention in Git commits.

> (It's minor, but I'm also personally not a fan of mixing the imperative style description which can be seen as "first-party" with the "third-party" "Args", "Returns", "Raises", etc...)

I just interpret those as plurals rather than descriptive since a function has multiple arguments, can return multiple values, and can raise more than one thing.   It's a little of a stretch but makes me feel better :-)

My personal preference would be to adopt Google Style and PEP-257 and prefer PEP-257 where they conflict (i.e. use imperative mood).

As a side curiosity do we have a mood standard among any other languages at NI?

thejcannon: Unsurprisingly, Google's [developer documentation style guide](https://developers.google.com/style/reference-verbs) states:

> When you're writing reference documentation for a method, phrase the main method description in terms of what the method does (gets, lists, creates, searches), rather than what the developer would use it to do (get, list, create, search).

I've also read a handful of posts from various people online (language-agnostic) basically saying "client documentation uses descriptive. Developer documentation (I.e. comments) uses imperative". I think this makes the most as it fits with the context of what's being described and to whom.

Ultimately, I think I'm inclined to leave it as "We follow Google's guide." which would mean advocating for descriptive style across the company. However also calling out, much like _which_ guide to follow that being consistent is ultimately the important thing. The tooling currently doesn't enforce any verb-mood, and I'll switch it back to simply enforce all conventions (minus verb mood).

neilvana: Sounds good to me.  I also buy the developer versus user argument.  I've never heard that before but it makes sense.

neilvana (approved):

rick-morrow (approved):

ni_python_styleguide/_cli.py
alejandro5042: Agreed, I love their explanation. Talk in terms of what the method can do for the user (gets), instead of what the program will do if you use it (get).

kcuzner (approved):

thejcannon:
This will fail PR due to what appears to be a bug in `pydocstyle`. I made an issue here: https://github.com/PyCQA/pydocstyle/issues/523

pakdev (approved):

PR #47, opened 2021-02-02
add readme and project links to package

mshafer-NI:
This should cause our existing readme to be used on https://pypi.org/project/ni-python-styleguide/ - based on the docs: https://python-poetry.org/docs/pyproject/#readme

Also adding "homepage" as the GH-pages site and "repository" as the base repo links

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,7 +7,6 @@ version = "0.1.4-alpha.0"
 description = "NI's internal and external Python linter rules and plugins"
 authors = ["NI <opensource@ni.com>"]
 readme = "README.md" # apply the repo readme to the package as well
-homepage = "https://ni.github.io/python-styleguide/"
 repository = "https://github.com/ni/python-styleguide"
 license = "MIT"
 include = ["ni_python_styleguide/config.toml"]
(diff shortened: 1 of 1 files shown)


mshafer-NI:
(Note that I am unable to push and test what this looks like on Pypi - even on test.pypi as "The user 'mshafer-ni' isn't allowed to upload to project 'ni-python-styleguide'. See https://test.pypi.org/help/#project-name for more information.", and getting poetry to publish under a slightly different name was proving to be a challenge.

irwand (approved):

pyproject.toml
```diff
@@ -6,6 +6,9 @@ name = "ni-python-styleguide"
 version = "0.1.4-alpha.0"
 description = "NI's internal and external Python linter rules and plugins"
 authors = ["NI <opensource@ni.com>"]
+readme = "README.md" # apply the repo readme to the package as well
+homepage = "https://ni.github.io/python-styleguide/"
```
thejcannon, line 10: I think the homepage and the repo should be the same. This would be the homepage for _the tool_.

thejcannon (approved):
👍

PR #66, opened 2021-09-08
Avoid index errors when input file is empty

m-akinc:
When we try to add suppressions in empty files, it results in index out of range errors. Really, it's a result of a linter error being reported on a non-existent line (i.e. line 1 of empty file). The _add_noqa_to_line function takes in the array of lines and modifies the contents, so the easiest way to avoid this problem is to populate the lines array with a single blank line when the file is empty. Also, the multiline checker has a similar problem, but we can avoid that easily by checking that the error.line is within range _before_ calling multiline_checker.in_multiline_string.

### DIFF
--- a/ni_python_styleguide/_acknowledge_existing_errors/__init__.py
+++ b/ni_python_styleguide/_acknowledge_existing_errors/__init__.py
@@ -73,6 +73,10 @@ def acknowledge_lint_errors(lint_errors):
     for bad_file, errors_in_file in lint_errors_by_file.items():
         path = pathlib.Path(bad_file)
         lines = path.read_text().splitlines(keepends=True)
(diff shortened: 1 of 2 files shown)


mshafer-NI:
Can we add these cases to the system tests?
https://github.com/ni/python-styleguide/tree/main/tests/test_cli/acknowledge_existing_errors_test_cases__snapshots

irwand (approved):

m-akinc:
> Can we add these cases to the system tests?
> https://github.com/ni/python-styleguide/tree/main/tests/test_cli/acknowledge_existing_errors_test_cases__snapshots

Added test for empty file

mshafer-NI (approved):
Looks good - thanks for adding the test

PR #67, opened 2021-09-10
Improve multiline statement handling

mshafer-NI:
Add support for putting the suppression after a multi-line statement that uses line continuation markers (`\`)


ni_python_styleguide/_acknowledge_existing_errors/__init__.py
```diff
@@ -25,32 +25,34 @@ def in_multiline_string(self, lineno):
 
     @staticmethod
     def _count_multiline_string_endings_in_line(line):
-        return line.count('"""') + line.count("'''")
+        return line.count('"""'), line.count("'''")
```
mshafer-NI, line 28: learned that having the other quote in a multiline string tripped this up, now it's handled.

ni_python_styleguide/_acknowledge_existing_errors/__init__.py
```diff
@@ -25,32 +25,34 @@ def in_multiline_string(self, lineno):
...
     old_line_ending = "\n" if line.endswith("\n") else ""
     line = line.rstrip("\n")
 
-    existing_suppression = re.search(r"noqa (?P<existing_suppresions>[\w\d]+\: [\w\W]+?) -", line)
-    if existing_suppression:
-        before = existing_suppression.groupdict()["existing_suppresions"]
-        if error_code not in before:
-            line = line.replace(before, before + f", {error_code}: {explanation}")
-    else:
+    if f"noqa {error_code}" not in line:
```
mshafer-NI, line 55: I experimented with changing the multi-output to be actually be like:
```
  # noqa {code1}, {code2}: {explanation 1}, {explanation 2}
```

But when I was looking at the output, I realized that this created distance between the second code  - both making it harder to parse, and increasing the size of the diff when fixing one of the codes.

Since this wasn't working as intended, and I've realized that this slightly more verbose output is better anyways, simplifying to just append (snapshots show now change to output)

ni_python_styleguide/_acknowledge_existing_errors/__init__.py
```diff
@@ -64,7 +66,7 @@ def acknowledge_lint_errors(lint_errors):
     """
     parsed_errors = map(_lint_errors_parser.parse, lint_errors)
     parsed_errors = filter(None, parsed_errors)
-    lint_errors_to_process = [error for error in parsed_errors if error not in EXCLUDED_ERRORS]
+    lint_errors_to_process = [error for error in parsed_errors if error.code not in EXCLUDED_ERRORS]
```
mshafer-NI, line 69: Initial tests had BLK100 show up, not sure how this was missed previously.

ni_python_styleguide/_acknowledge_existing_errors/__init__.py
```diff
@@ -93,7 +95,7 @@ def acknowledge_lint_errors(lint_errors):
             cached_key = f"{error.file}:{error.line + skip}"
             if error.code in handled_lines[cached_key]:
                 logging.warning(
-                    "Multiple occurances of error %s code were logged for %s:%s, only suprressing first",
+                    "Multiple occurrences of error %s code were logged for %s:%s, only suprressing first",
```
mshafer-NI, line 98: Thank you spell checker

irwand (approved):

PR #72, opened 2021-10-28
Better handle when acknowledging violations will cause reformat

mshafer-NI:
Adding a "--aggressive" flag to `acknowledge-existing-violations` that will attempt to suppres, then format, then move suppressions back to the correct lines, and format, and repeat until formatting no longer changes the file (max of 10 tries).

1. Extracted running the linter to `_lint.lint`
1. Extracted config file consts from `_cli.py` to `_config_constants.py`
1. Create `_format.formt` helper method for running `black` (decided to leave adding potential "format" command to future PR)
1. Split out the per-file code from `_acknowledge_existing_errors.acknowledge_lint_errors` to `_acknowledge_existing_errors._suppress_errors_in_file`
1. Add conditional `handle_lines_that_are_now_too_long` (this re-uses _suppress_errors_in_file for subsequent runs).
1. Added test that uses "--agressive" flag and set output files to not be ignored by black (by changing the suffix so it didn't match the existing filter).

Example run:
(starting code)
```python
def method_withBadName_andParams(my_normal_param, myBadlyNamedParam, my_other_Bad_param): 
    """Provide example where black will want to split out result."""
    return 5 + 7

```

1.
This contains several errors, so round 1 will suppress those:
```python
def method_withBadName_andParams(my_normal_param, myBadlyNamedParam, my_other_Bad_param):  # noqa N802: function name 'method_withBadName_andParams' should be lowercase (auto-generated noqa)  # noqa N803: argument name 'myBadlyNamedParam' should be lowercase (auto-generated noqa)
    """Provide example where black will want to split out result."""
    return 5 + 7

```

Now the line is too long, so when `black` is called, it will break up the line:

```python
def method_withBadName_andParams(
    my_normal_param, myBadlyNamedParam, my_other_Bad_param
):  # noqa N802: function name 'method_withBadName_andParams' should be lowercase (auto-generated noqa)  # noqa N803: argument name 'myBadlyNamedParam' should be lowercase (auto-generated noqa)
    """Provide example where black will want to split out result."""
    return 5 + 7

```

However, now the suppressions are on the wrong line.
Because the file was reformatted - we attempt again.

2.
We remove the suppressions and re-lint and suppress:

```python
def method_withBadName_andParams(  # noqa N802: function name 'method_withBadName_andParams' should be lowercase (auto-generated noqa)
    my_normal_param, myBadlyNamedParam, my_other_Bad_param  # noqa N803: argument name 'myBadlyNamedParam' should be lowercase (auto-generated noqa)
): 
    """Provide example where black will want to split out result."""
    return 5 + 7
```

The args line is still too long, so `black` will again format
```python
def method_withBadName_andParams(  # noqa N802: function name 'method_withBadName_andParams' should be lowercase (auto-generated noqa)
    my_normal_param, 
    myBadlyNamedParam, 
    my_other_Bad_param,  # noqa N803: argument name 'myBadlyNamedParam' should be lowercase (auto-generated noqa)
): 
    """Provide example where black will want to split out result."""
    return 5 + 7
```

Again, now the suppressions are on the wrong line, but we know it was reformatted, so we attempt again.

3.
```python
def method_withBadName_andParams(  # noqa N802: function name 'method_withBadName_andParams' should be lowercase (auto-generated noqa)
    my_normal_param, 
    myBadlyNamedParam,  # noqa N803: argument name 'myBadlyNamedParam' should be lowercase (auto-generated noqa)
    my_other_Bad_param,
): 
    """Provide example where black will want to split out result."""
    return 5 + 7
```

And this time, `black` is happy with the output, and so doesn't format and leaves the suppressions on the correct lines.


ni_python_styleguide/_acknowledge_existing_errors/__init__.py
```diff
@@ -59,57 +72,170 @@ def _add_noqa_to_line(lineno, code_lines, error_code, explanation):
     code_lines[lineno] = line + old_line_ending
 
 
-def acknowledge_lint_errors(lint_errors):
+class _Acknowlegder:
```
irwand, line 75: I don't get the essence or mechanism of this class.. seems like a hodge-podge of miscellaneous helper being dumped here, which makes it harder to discern the flow of the logic to me. I do like that the methods have clear names.

irwand: is there specific reasons why the class is needed? seems better to me without the class maybe?

ni_python_styleguide/_acknowledge_existing_errors/__init__.py
```diff
@@ -59,57 +72,170 @@ def _add_noqa_to_line(lineno, code_lines, error_code, explanation):
...
+
+        # format - are we done?
+        done = self.run_formatter(file)
+        if not done:
+            try:
+                limit_ = limit
+                if limit is None:
+                    limit_ = _Acknowlegder.MAX_FORMAT_RETRIES_LIMIT
+
+                self.handle_lines_that_are_now_too_long(file, limit=limit_ - 1)
```
irwand, line 159: seems like there's no reason why this needs to be recursive. Making it a loop IMO make it more readable.

mshafer-NI: changed to loop (for _ in range(limit) / if done: break / else raise)

ni_python_styleguide/_acknowledge_existing_errors/__init__.py
mshafer-NI: to hold the 4 linter args - changed to using funtools.partial for the subsequent calls

ni_python_styleguide/_acknowledge_existing_errors/__init__.py
```diff
@@ -1,16 +1,29 @@
...
 from collections import defaultdict
+from collections.abc import Iterable
+from io import StringIO
+from typing import Union
 
+from ni_python_styleguide import _format
+from ni_python_styleguide import _lint
 from ni_python_styleguide._acknowledge_existing_errors import _lint_errors_parser
 
+MODULE_LOGGER = logging.getLogger(__name__)
```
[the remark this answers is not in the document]
mshafer-NI: it should never get re-assigned??

mshafer-NI:
Note: I attempted to move the "--aggressive" logic up a level to the CLI (such that if --aggressive is passed, it would find all included .py files, call down to remove suppressions, call down to suppress, then do the loop.) however, this yielded a lot of difficulties. I will probably re-attempt when I next get a chance work on this (unless you disagree)

ni_python_styleguide/_format.py
```diff
@@ -0,0 +1,35 @@
...
+            ]
+        )
+    except SystemExit as e:
+        if e.code == 0:
+            pass  # why are they exiting when called via import ?! ‾\_(ツ)_/‾
+        else:
+            raise
+
+
+def does_formatting_make_changes(file_or_dir):
```
irwand, line 27: can this just be `format_check` ?

irwand (approved):
code easier to understand IMO. Thx!

ni_python_styleguide/_format.py
mshafer-NI: probably - do you think it should just do a "check" in that case? (It's currently doing a format)

mshafer-NI: oh, nope we have to do a format due to the pydocstyle bug :(

ni_python_styleguide/_acknowledge_existing_errors/__init__.py
```diff
@@ -54,62 +66,125 @@ def _add_noqa_to_line(lineno, code_lines, error_code, explanation):
...
+):
     """Add a "noqa" comment for each of existing errors (unless excluded).
 
     Excluded error (reason):
     BLK100 - run black
     """
-    parsed_errors = map(_lint_errors_parser.parse, lint_errors)
-    parsed_errors = filter(None, parsed_errors)
-    lint_errors_to_process = [error for error in parsed_errors if error.code not in EXCLUDED_ERRORS]
+    get_lint_errors_to_process__simple = functools.partial(
```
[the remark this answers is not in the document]
mshafer-NI: removed the partial

ni_python_styleguide/_acknowledge_existing_errors/__init__.py
```diff
@@ -54,62 +66,125 @@ def _add_noqa_to_line(lineno, code_lines, error_code, explanation):
...
+                # re-apply suppressions on correct lines
+                _remove_auto_suppressions_from_file(bad_file)
+                current_lint_errors = get_lint_errors_to_process__simple([bad_file])
+                _suppress_errors_in_file(bad_file, current_lint_errors, encoding=DEFAULT_ENCODING)
+
+                changed = _format.does_formatting_make_changes(bad_file)
+                if not changed:  # are we done?
+                    break
+            else:
+                raise LimitReachedError(
```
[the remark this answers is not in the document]
mshafer-NI: Changed to RunTimeError

ni_python_styleguide/_format.py
mshafer-NI: changed method name to format_check

mshafer-NI:
> Note: I attempted to move the "--aggressive" logic up a level to the CLI (such that if --aggressive is passed, it would find all included .py files, call down to remove suppressions, call down to suppress, then do the loop.) however, this yielded a lot of difficulties. I will probably re-attempt when I next get a chance work on this (unless you disagree)

This is already a pretty big diff. I am opting to leave it as is and will come in later to add a "--remove-current-acknowledgements" arg

mshafer-NI:
Blocked on #73

ni_python_styleguide/_acknowledge_existing_errors/__init__.py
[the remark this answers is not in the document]
mshafer-NI: changed to _module_logger. PR to update docs to explicitly handle this case should be coming soon.

PR #75, opened 2021-11-17
treat global mutables as global vars

mshafer-NI:
Add clarification that if a global is mutable, to treat is as a global var (snake_case) as opposed to a "constant reference" (SCREAMING_CASE)

### DIFF
--- a/docs/Coding-Conventions.md
+++ b/docs/Coding-Conventions.md
@@ -340,6 +340,20 @@ FlyingCircus_contra = TypeVar("FlyingCircus_contra", contravariant=True)
 
 ℹ️ This invokes Python's name mangling which does have well-known, yet unintended side-effects. See [the docs](https://docs.python.org/3.6/tutorial/classes.html#private-variables)
 
+### [N.2.12] ✔️ **DO** Name global *mutable* objects as global variables (i.e., use `snake_case`)
+
+ℹ️ While it can be argued that a global reference that should not be re-assigned is a CONST, we chose to acknowledge that global *mutable* objects are variables whose behavior may change without getting re-assigned.
+
+```python
+# Bad
+CHEESE_LOGGER = logging.getLogger("Cheese")
+```
+
+```python
+# Good
+_logger = logging.getLogger("Cheese")
+```
+
 ---
 
 # [L] Language Features


irwand (approved):

mshafer-NI:
comment job is broken...

Since this changes the Coding Conventions, I'll @-mention the appropriate NI engineers.

@alejandro5042  @alexweav  @DavidCurtiss  @irwand  @jryckman  @kcuzner  @mshafer-NI  @neilvana  @nethrasuresh
@pakdev @rick-morrow  @rtzoeller  @sbethur @stick152 @Adithyak1998 @innagarc @ShibaniRout

DavidCurtiss (approved):

rtzoeller (approved):

rick-morrow (approved):

docs/Coding-Conventions.md
```diff
@@ -340,6 +340,20 @@ FlyingCircus_contra = TypeVar("FlyingCircus_contra", contravariant=True)
 
 ℹ️ This invokes Python's name mangling which does have well-known, yet unintended side-effects. See [the docs](https://docs.python.org/3.6/tutorial/classes.html#private-variables)
 
+### [N.2.12] ✔️ **DO** Name global *mutable* objects as global variables (i.e., use `snake_case`)
```
mshafer-NI, line 343: @kcuzner 
> Is this a common convention for python? In rust mutable `'static` globals are still SCREAMING_SNAKE_CASE.

I can't speak to many languages, but I do know that for:

* C++ - something being passed as const reference (`const Foo& foo`) (which is a perfect analogy as we're saying don't change the reference - see more [here](https://www.fluentcpp.com/2018/07/13/the-incredible-const-reference-that-isnt-const/#:~:text=Const%20reference%20or%20reference%20to%20const%3F)) results in skipping the copy, but also not allowing the receiving method to change the reference, but the object pointed at may be mutated.
* C# - only lets you call something `const` if it's compile-time constant, it does offer the `readonly` keyword for "at execution time, create this once"; but because it is common to configure the logger within the main block of a script (or the arg-parsing block...), this isn't an accurate parallel (unless we view it as a variable)

mshafer-NI: Google styleguide doesn't say one way or other, but their one GLOBAL example is immutable.... (https://google.github.io/styleguide/pyguide.html#254-decision)

kcuzner (approved):

docs/Coding-Conventions.md
kcuzner: I think its useful to know that the variable being mutated is global and `SCREAMING_CASE` makes that obvious to me. However, if the reigning consensus is that `SCREAMING_CASE` === `const` I can get on board with having mutable globals `snake_case` since it makes it obvious when you're modifying something that wasn't meant to be modified even if it reduces how obvious the scope of the variable being changed is..

neilvana (approved):

PR #77, opened 2022-02-02
Be more lenient on dependency versioning by using >=

irwand:
This is to help reduce problems with `poetry lock` due to over-restricted dependency versioning in client's python modules.


mshafer-NI:
weird, it looks like this is completely breaking the auto-acknowledge code...

mshafer-NI (approved):

neilvana (approved):

PR #80, opened 2022-03-31
Add fix command

mshafer-NI:
Create "ni-python-styleguide fix" command to:
* sort imports (uses isort, but then splits imports to one per line)
* format with black
* if --aggressive is passed, remove suppressions, first, then re-acknowledge anything remaining.


ni_python_styleguide/_acknowledge_existing_errors/__init__.py
```diff
@@ -13,45 +14,6 @@
     "BLK100",
 }
 
-DEFAULT_ENCODING = "UTF-8"
```
mshafer-NI, line 16: moved to _utils.string_helpers

ni_python_styleguide/_acknowledge_existing_errors/__init__.py
```diff
@@ -108,42 +74,64 @@ def acknowledge_lint_errors(
 
     failed_files = []
     for bad_file, errors_in_file in lint_errors_by_file.items():
-        _suppress_errors_in_file(bad_file, errors_in_file, encoding=DEFAULT_ENCODING)
```
mshafer-NI, line 111: thinking about this some more, I think it would be better to merge this back into current entry point and have _fix call with the file in question wrapped in a list

ni_python_styleguide/_cli.py
```diff
@@ -158,3 +161,27 @@ def acknowledge_existing_violations(obj, extend_ignore, file_or_dir, aggressive)
...
+@click.option(
+    "--extend-ignore",
+    type=str,
+    help="Comma-separated list of errors and warnings to ignore (or skip)",
+)
+@click.argument("file_or_dir", nargs=-1)
+@click.option(
+    "--aggressive",
+    is_flag=True,
+    help="Remove any existing suppressions, fix what can be fixed, and suppress remaining.",
```
mshafer-NI, line 176: s/suppress/acknowledge/

ni_python_styleguide/_fix.py
```diff
@@ -0,0 +1,159 @@
...
+
+def fix(exclude, app_import_names, extend_ignore, file_or_dir, *_, aggressive=False, diff=False):
+    """Fix basic linter errors and format."""
+    file_or_dir = file_or_dir or ["."]
+    if aggressive:
+        all_files = []
+        for file_or_dir_ in file_or_dir:
+            file_path = pathlib.Path(file_or_dir_)
+            if file_path.is_dir():
+                if _utils.git.find_repo_root(file_or_dir_):
```
irwand, line 101: I really dislike coupling this with git at all.. the tool should not even try to see if there's git.. it should blindly take the input file/dir, and iterate through.. is there compelling reason why coupling this to git is desirable?

ni_python_styleguide/_acknowledge_existing_errors/__init__.py
```diff
@@ -13,45 +14,6 @@
...
-
-            # if occurrences of multiline string markers is odd, this must be in a multiline
-            #  or, if line continuation token is on the ending, assume in a multiline statement
-            self._values.append(
-                any([part % 2 == 1 for part in current_count])
-                or code_part_of_line.strip().endswith("\\")
-            )
-
 
 def _add_noqa_to_line(lineno, code_lines, error_code, explanation):
```
irwand, line 18: should the rest of the helper be moved to _utils?

ni_python_styleguide/_fix.py
mshafer-NI: My thought was that if "--aggressive", that we should remove any existing acknowledge statements from all relevant files, then put back what was still needed.

I could see making it only `{path}/**.py`; however, in the event that a user calls this where venv's would match, that will (a) make it take a long time, and (b) make it try to lint/format a LOT of files that shouldn't be.

ni_python_styleguide/_acknowledge_existing_errors/__init__.py
mshafer-NI: the rest of `_get_lint_errors_to_process`?, should probably move to `_lint` as `git_lint_errors`...

ni_python_styleguide/_fix.py
irwand: That's up to the user to figure out.. they said: process-this-directory, we should say "yes sir!" .. simpler mental model and predictable is superior than trying to be smart.

ni_python_styleguide/_acknowledge_existing_errors/__init__.py
irwand: maybe.. I could see using __init__ to contain some stuff as helper.. but now you have _utils to contain those things.. we should then move things over.

ni_python_styleguide/_fix.py
mshafer-NI: Changed to an rglob.

ni_python_styleguide/_acknowledge_existing_errors/__init__.py
mshafer-NI: moved this refactor to #82

ni_python_styleguide/_cli.py
```diff
@@ -158,3 +161,27 @@ def acknowledge_existing_violations(obj, extend_ignore, file_or_dir, aggressive)
...
+@click.option(
+    "--extend-ignore",
+    type=str,
+    help="Comma-separated list of errors and warnings to ignore (or skip)",
+)
+@click.argument("file_or_dir", nargs=-1)
+@click.option(
+    "--aggressive",
+    is_flag=True,
+    help="Remove any existing acknowledgments, fix what can be fixed, and re-acknowledges remaining.",
```
mshafer-NI, line 176: re-acknowledge

ni_python_styleguide/_fix.py
```diff
@@ -0,0 +1,166 @@
...
+
+    >>> _split_imports_line("import os, collections\nimport pathlib, os")
+    'import os\nimport collections\nimport pathlib\nimport os\n'
+
+    >>> _split_imports_line("\n")
+    '\n'
+    """  # noqa W505: long lines...
+    result_parts = []
+    for line in lines.splitlines(keepends=True):
+        code_portion_of_line, *non_code = line.split("#", maxsplit=1)
```
mshafer-NI, line 44: does this need to pack?

ni_python_styleguide/_fix.py
```diff
@@ -0,0 +1,166 @@
...
+
+
+def fix(
+    exclude: str,
+    app_import_names: str,
+    extend_ignore,
+    file_or_dir,
+    *_,
+    aggressive=False,
+    diff=False,
```
mshafer-NI, line 100: not handled yet - remove for now

ni_python_styleguide/_fix.py
mshafer-NI: yes it does

irwand (approved):

mshafer-NI:
Included in v0.1.11 -> https://github.com/ni/python-styleguide/releases/tag/v0.1.11

PR #87, opened 2022-09-08
Use black profile with isort to reduce manual work

mshafer-NI:
I've learned that `isort` now has a `black` profile allowing it to play nicely with `black`.

To start investigating bringing `flake8-isort` in, I opted to first test what configurations we would need to match our current style.

Changing to configuring the call to `isort` within `_sort_imports` removes the need for the custom logic around splitting multi-import lines, so choosing to contribute this simplification at this point.

### DIFF
--- a/tests/test_cli/fix_test_cases__snapshots/basic_example/output.py
+++ b/tests/test_cli/fix_test_cases__snapshots/basic_example/output.py
@@ -2,9 +2,9 @@
 import pathlib
 from os import access
 from os import path
-from typing import (
+from typing import (  # noqa F401: un-used import comment that is actually used, should get removed in --aggressive (auto-generated noqa)
     Hashable,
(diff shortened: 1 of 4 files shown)


tests/test_cli/fix_test_cases__snapshots/basic_example/output.py
```diff
@@ -2,9 +2,9 @@
 import pathlib
 from os import access
 from os import path
-from typing import (
+from typing import (  # noqa F401: un-used import comment that is actually used, should get removed in --aggressive (auto-generated noqa)
```
mshafer-NI, line 5: our custom logic moved this to after the paren - this is an unused comment and does not need to be here (`--aggresive` mode still removes it)

irwand (approved):

PR #95, opened 2022-12-08
Disallow black 22.10.0 because the initial macOS wheels had the wrong architecture

bkeryan:
When `black` 22.10.0 was released, the initial macOS x86_64 wheels were actually arm64 wheels: https://github.com/psf/black/issues/3312

Apparently this has been corrected, but some of our GitHub agents still have the original, incorrect wheels in their Poetry caches.

Disallow `black` 22.10.0 to ensure that clients of `ni-python-styleguide` don't run into this problem.

### DIFF
--- a/.github/workflows/PR.yml
+++ b/.github/workflows/PR.yml
@@ -50,7 +50,7 @@ jobs:
     strategy:
       matrix:
         os: [macos-latest, windows-latest, ubuntu-latest]
-        python-version: [3.6, 3.7, 3.8, 3.9]
+        python-version: [3.7, 3.8, 3.9]
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
--- a/poetry.lock
+++ b/poetry.lock
@@ -1,12 +1,4 @@
 [[package]]
-name = "atomicwrites"
-version = "1.4.1"
-description = "Atomic file writes."
-category = "dev"
-optional = false
-python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*"
-
-[[package]]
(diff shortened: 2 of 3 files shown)


bkeryan:
```
Run actions/setup-python@v2
Version 3.6 was not found in the local cache
Error: Version 3.6 with arch x64 not found
The list of all available versions can be found here: https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json
```

mshafer-NI:
> ```
> Run actions/setup-python@v2
> Version 3.6 was not found in the local cache
> Error: Version 3.6 with arch x64 not found
> The list of all available versions can be found here: https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json
> ```

‾\_(ツ)_/‾, 3.6 has been eol for a while. I guess it's time to stop testing it here...

rtzoeller:
> the initial macOS x86_64 wheels were actually amd64 wheels

Typo? I think it should be arm64 or aarch64.

.github/workflows/PR.yml
```diff
@@ -50,7 +50,7 @@ jobs:
     strategy:
       matrix:
         os: [macos-latest, windows-latest, ubuntu-latest]
-        python-version: [3.6, 3.7, 3.8, 3.9]
+        python-version: [3.7, 3.8, 3.9]
```
bkeryan, line 53: We should commit https://github.com/ni/python-styleguide/pull/96 first so that removing 3.6 here is no longer part of this PR.

irwand: I committed that other one. Thx !!

irwand (approved):

PR #99, opened 2023-01-09
Fix up tooling

mshafer-NI:
* Upgrade the publish pipeline to use a modern tooling stack (based on another NI repo that is successfully using it)
* Attempt to get PR comment to only run once (appears that if this works, it needs to be in the base branch...)
* Update to using poetry 1.2.2 for checking if lock is up to date
* Upgrade Gr1n/setup-poetry to latest version that (a) removes deprecated comment from GitHub, and (b) adds support for newer Python versions
* Add now tooling-supported newer Python versions to tests
* Create weekly action that should:
  * Run poetry lock
  * reformat if needed
  * post PR


.github/workflows/Comment-on-PR.yml
```diff
@@ -1,8 +1,11 @@
 on:
   pull_request_target:
     paths:
-      - '*'
-      - '!docs/Coding-Conventions.md'
+      - "*"
+      - "!docs/Coding-Conventions.md"
+    types:
```
mshafer-NI, line 6: :frowning:, reading the docs (https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target), I'm pretty sure this is right, but it's not triggering on thie PR....

irwand (approved):

.github/workflows/Comment-on-PR.yml
mshafer-NI: I think I want to go ahead and merge this to keep iterating

PR #101, opened 2023-01-19
Fix comments

mshafer-NI:
#99 appears to have broken the comment on PRs, this fixes it.

Testing by opening PR in fork: https://github.com/mshafer-NI/ni-python-styleguide/pull/23


mshafer-NI:
:sigh: I got a comment in my fork, so either this works if already merged, or we'll have to go back to not specifying triggers and getting a comment on every update as wel...

irwand (approved):

PR #102, opened 2023-01-19
Fix no split submodule imports

### LINKED ISSUE #97: The `lint` and `fix` subcommands disagree about whether to split imports
The `lint` subcommand does not warn about combined imports like `from os import path, access`, but the `fix` subcommand splits them into multiple lines anyway. However, it only does this when there are lint warnings. If there are no lint warnings, `fix` leaves the combined imports alone.

I suggest configuring `fix` to use/preserve combined imports, because they are more concise and I don't think they hurt readability. However, it would also be reasonable to update `lint` to warn about combined imports.

### An example from ni-measurementlink-service
```
# before
from typing import Iterable, List, NamedTuple, Optional

from ni_measurementlink_service._internal.stubs.ni.measurementlink.sessionmanagement.v1 import (
    session_management_service_pb2,
    session_management_service_pb2_grpc,
)
```
```
# after
from typing import Iterable
from typing import List
from typing import NamedTuple
from typing import Optional

from ni_measurementlink_service._internal.stubs.ni.measurementlink.sessionmanagement.v1 import (
    session_management_service_pb2,
)
from ni_measurementlink_service._internal.stubs.ni.measurementlink.sessionmanagement.v1 import (
    session_management_service_pb2_grpc,
)
```

### Steps to reproduce

`tests\test_cli\fix_test_cases__snapshots\basic_example\input.py` demonstrates the behavior.

1. Run `copy .\tests\test_cli\fix_test_cases__snapshots\basic_example\input.py .`
2. Run `poetry run ni-python-styleguide lint input.py`. None of the warnings are about combined imports:
```
PS D:\dev\python-styleguide> poetry run ni-python-styleguide lint input.py
input.py:7:1: I201 Missing newline between import groups. 'import pytest' is identified as Third Party and 'from typing import Iterable, List, Hashable' is identified as Stdlib.
input.py:8:1: I100 Import statements are in the wrong order. 'import pathlib' should be before 'import pytest' and in a different group.
input.py:8:1: I201 Missing newline between import groups. 'import pathlib' is identified as Stdlib and 'import pytest' is identified as Third Party.
```
3. Manually fix the warnings in `input.py`:
```
import pathlib
from os import path, access
from typing import (
    Iterable,
    List,
    Hashable,  # noqa F401: un-used import comment that is actually used, should get removed in --aggressive (auto-generated noqa)
)

import pytest
```
4. Run `poetry run ni-python-styleguide lint input.py`. There are no warnings.
6. Run `poetry run ni-python-styleguide fix input.py`. It does not make any changes.
7. Run `copy .\tests\test_cli\fix_test_cases__snapshots\basic_example\input.py input2.py`
8. Run `poetry run ni-python-styleguide fix input2.py`. It splits the imports:
```
PS D:\dev\python-styleguide> poetry run ni-python-styleguide fix input2.py
All done! ✨ 🍰 ✨
1 file left unchanged.
All done! ✨ 🍰 ✨
1 file left unchanged.
WARNING:flake8.options.manager:option --indent-size: please update `help=` text to use %(default)s instead of %default -- this will be an error in the future
WARNING:flake8.options.manager:option --max-complexity: please update from optparse string `type=` to argparse callable `type=` -- this will be an error in the future
WARNING:flake8.checker:The multiprocessing module is not available. Ignoring --jobs arguments.
```
```
import pathlib
from os import access
from os import path
from typing import (  # noqa F401: un-used import comment that is actually used, should get removed in --aggressive (auto-generated noqa)
    Hashable,
)
from typing import Iterable
from typing import List

import pytest
```

mshafer-NI:
Fixes #97 

Based on a poll with NI engineers, we are choosing to align on ["Do not put module level imports on the same line"](https://ni.github.io/python-styleguide/#O-1-1), but allowing sub-module imports on the same line.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,7 +3,7 @@ name = "ni-python-styleguide"
 # The -alpha.0 here denotes a source based version
 # This is removed when released through the Publish-Package.yml GitHub action
 # Official PyPI releases follow Major.Minor.Patch
-version = "0.3.0-alpha.0"
+version = "0.4.0-alpha.0"
 description = "NI's internal and external Python linter rules and plugins"
 authors = ["NI <opensource@ni.com>"]
 readme = "README.md" # apply the repo readme to the package as well
--- a/ni_python_styleguide/config.toml
+++ b/ni_python_styleguide/config.toml
@@ -8,4 +8,3 @@ line-length = 100
 [tool.isort]
 profile = "black"
 combine_as_imports = true
-force_single_line = true
--- a/tests/test_cli/fix_test_cases__snapshots/basic_example/output.py
+++ b/tests/test_cli/fix_test_cases__snapshots/basic_example/output.py
@@ -1,12 +1,10 @@
 """Provide example cases of imports that need sorting and a file that needs formatted."""
 import pathlib
-from os import access
-from os import path
+from os import access, path
 from typing import (  # noqa F401: un-used import comment that is actually used, should get removed in --aggressive (auto-generated noqa)
     Hashable,
 )
-from typing import Iterable
-from typing import List
+from typing import Iterable, List
 
 import pytest
 
--- a/tests/test_cli/fix_test_cases__snapshots/basic_example/output__aggressive.py
+++ b/tests/test_cli/fix_test_cases__snapshots/basic_example/output__aggressive.py
@@ -1,10 +1,7 @@
 """Provide example cases of imports that need sorting and a file that needs formatted."""
 import pathlib
-from os import access
-from os import path
-from typing import Hashable
-from typing import Iterable
-from typing import List
+from os import access, path
+from typing import Hashable, Iterable, List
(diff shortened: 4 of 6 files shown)


pyproject.toml
```diff
@@ -3,7 +3,7 @@ name = "ni-python-styleguide"
 # The -alpha.0 here denotes a source based version
 # This is removed when released through the Publish-Package.yml GitHub action
 # Official PyPI releases follow Major.Minor.Patch
-version = "0.3.0-alpha.0"
+version = "0.4.0-alpha.0"
```
mshafer-NI, line 6: the release flow currently doesn't do the version bump, so doing it manually: https://pypi.org/project/ni-python-styleguide/0.3.0/

bkeryan (approved):

irwand (approved):

PR #114, opened 2023-03-15
Allow running ni-python-styleguide as a pre-commit hook

bkeryan:

Add a [pre-commit](https://[pre-commit](https://pre-commit.com/).com/) hooks file so that other projects can run `ni-python-styleguide` as a pre-commit hook.

Add a `pre-commit` config file to enable some pre-commit hooks for commits to this repo.

Running `ni-python-styleguide` automatically may save developers time and prevent failed PR builds. 

Tested running `ni-python-styleguide` with `poetry1.2 run pre-commit try-repo C:\dev\python-styleguide ni-python-styleguide --all-files`.
Tested this repo's hooks with `poetry run pre-commit run --all-files`.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,11 @@
+repos:
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.4.0
+    hooks:
+    -   id: check-toml
+    -   id: check-yaml
+-   repo: https://github.com/psf/black
+    rev: 23.1.0
+    hooks:
+    -   id: black
+        exclude: '^tests/test_cli/acknowledge_existing_errors_test_cases__snapshots/'
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -5,4 +5,3 @@
     language: python
     types: [python]
     args: [lint]
-    pass_filenames: false
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,0 +1,8 @@
+-   id: ni-python-styleguide
+    name: ni-python-styleguide
+    description: Enforces NI's internal and external Python conventions.
+    entry: ni-python-styleguide
+    language: python
+    types: [python]
+    args: [lint]
+    pass_filenames: false
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -23,6 +23,16 @@ Please follow these steps to have your contribution considered by the maintainer
 
 ## ⭐️ Common Contributing Workflows ⭐️
 
+## Install Git `pre-commit` Hook Scripts
+
+This project uses [pre-commit](https://pre-commit.com/) to automatically run checks
+when you commit changes to the code. Install the `pre-commit` hook scripts in
+your local copy of the repo by running:
+
+```cmd
+poetry run pre-commit install
+```
+
 ### Adding a new plugin
 
 If you want to add a plugin to the list of plugins, here's how:
--- a/README.md
+++ b/README.md
@@ -69,3 +69,20 @@ line-length = 100
 ### Editor Integration
 
 (This section to come!)
+
+### Git `pre-commit` Hook Integration
+
+To use [pre-commit](https://pre-commit.com/) to automatically run `ni-python-styleguide` when you commit changes:
+- Add a dev-dependency on `pre-commit`.
+- Add this to your repo's `.pre-commit-config.yaml`:
+
+```yaml
+repos:
+-   repo: https://github.com/ni/python-styleguide/
+    rev: v0.4.1
+    hooks:
+    -   id: ni-python-styleguide
+```
+
+- Install the `pre-commit` hook scripts in your local copy of the repo by running `pre-commit install`. If your
+  project uses Poetry, run `poetry run pre-commit install` to use the virtualenv managed by Poetry.
--- a/poetry.lock
+++ b/poetry.lock
@@ -39,6 +39,14 @@ jupyter = ["ipython (>=7.8.0)", "tokenize-rt (>=3.2.0)"]
 uvloop = ["uvloop (>=0.15.2)"]
 
 [[package]]
+name = "cfgv"
+version = "3.3.1"
+description = "Validate configuration and produce human readable error messages."
+category = "dev"
+optional = false
+python-versions = ">=3.6.1"
+
+[[package]]
 name = "click"
 version = "8.1.3"
 description = "Composable command line interface toolkit"
@@ -59,8 +67,16 @@ optional = false
 python-versions = "!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*,!=3.4.*,!=3.5.*,!=3.6.*,>=2.7"
 
 [[package]]
+name = "distlib"
+version = "0.3.6"
+description = "Distribution utilities"
+category = "dev"
+optional = false
+python-versions = "*"
+
+[[package]]
 name = "exceptiongroup"
-version = "1.1.0"
+version = "1.1.1"
 description = "Backport of PEP 654 (exception groups)"
 category = "dev"
 optional = false
@@ -68,6 +84,18 @@ python-versions = ">=3.7"
 
 [package.extras]
 test = ["pytest (>=6)"]
+
+[[package]]
+name = "filelock"
+version = "3.10.0"
+description = "A platform independent file lock."
+category = "dev"
+optional = false
+python-versions = ">=3.7"
+
+[package.extras]
+docs = ["furo (>=2022.12.7)", "sphinx (>=6.1.3)", "sphinx-autodoc-typehints (>=1.22,!=1.23.4)"]
+testing = ["covdefaults (>=2.3)", "coverage (>=7.2.1)", "pytest (>=7.2.2)", "pytest-cov (>=4)", "pytest-timeout (>=2.1)"]
 
 [[package]]
 name = "flake8"
@@ -121,6 +149,17 @@ python-versions = "*"
 
 [package.dependencies]
 pycodestyle = "*"
+
+[[package]]
+name = "identify"
+version = "2.5.20"
+description = "File identification library for Python"
+category = "dev"
+optional = false
+python-versions = ">=3.7"
+
+[package.extras]
+license = ["ukkonen"]
 
 [[package]]
 name = "importlib-metadata"
@@ -177,6 +216,14 @@ optional = false
 python-versions = ">=3.5"
 
 [[package]]
+name = "nodeenv"
+version = "1.7.0"
+description = "Node.js virtual environment builder"
+category = "dev"
+optional = false
+python-versions = ">=2.7,!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*,!=3.4.*,!=3.5.*,!=3.6.*"
+
+[[package]]
 name = "packaging"
 version = "23.0"
 description = "Core utilities for Python packages"
@@ -186,7 +233,7 @@ python-versions = ">=3.7"
 
 [[package]]
 name = "pathspec"
-version = "0.11.0"
+version = "0.11.1"
 description = "Utility library for gitignore style pattern matching of file paths."
 category = "main"
 optional = false
@@ -205,7 +252,7 @@ flake8 = ">=5.0.0"
 
 [[package]]
 name = "platformdirs"
-version = "3.1.1"
+version = "2.6.2"
 description = "A small Python package for determining appropriate platform-specific dirs, e.g. a \"user data dir\"."
 category = "main"
 optional = false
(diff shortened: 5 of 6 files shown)


bkeryan:
This test is failing without my change:

```
====================================================== FAILURES =======================================================
________________________________ test_bad_codeblocks_document_lint_errors[D.1.12-Bad1] ________________________________

lint_codeblock = <function lint_codeblock.<locals>.run_linter at 0x000001D5DBA21EE0>
bad_codeblock = <tests.test_convention_doc.doctypes.Codeblock object at 0x000001D5DA59FD30>

    def test_bad_codeblocks_document_lint_errors(lint_codeblock, bad_codeblock):
        """Tests that "bad" codeblocks fail to lint and documents the resulting error codes."""
        if bad_codeblock.rule.is_automatically_enforced:
            result = lint_codeblock(
                bad_codeblock,
                "--format='%(code)s'",
                ignore_unused_imports=("F401" not in bad_codeblock.rule.error_codes),
            )
            assert not result, result.output

            error_codes = set(code.strip("'") for code in result.output.splitlines())
            assert error_codes
            expected_error_codes = set(bad_codeblock.rule.error_codes)
            error_codes_not_expected = error_codes - expected_error_codes
>           assert error_codes.issubset(
                expected_error_codes
            ), f'"Bad" codeblock caused unexpected lint error - {error_codes_not_expected}'
E           AssertionError: "Bad" codeblock caused unexpected lint error - {'BLK100'}
E           assert False
E            +  where False = <built-in method issubset of set object at 0x000001D5DB8E3E40>({'D201', 'D211'})
E            +    where <built-in method issubset of set object at 0x000001D5DB8E3E40> = {'BLK100', 'D201'}.issubset

tests\test_convention_doc\test_codeblocks.py:68: AssertionError
-------------------------------------------------- Captured log call --------------------------------------------------
INFO     flake8:flake8_black.py:63 flake8-black: loading black settings from C:\Dev\python-styleguide\ni_python_styleguide\config.toml
WARNING  flake8.plugins.reporter:reporter.py:35 "'%(code)s'" is an unknown formatter.  Falling back to default.
WARNING  flake8.checker:checker.py:105 The multiprocessing module is not available. Ignoring --jobs arguments.
INFO     flake8.checker:checker.py:264 Making checkers
INFO     flake8.checker:checker.py:175 Checking 1 files
INFO     flake8.main.application:application.py:234 Finished running
INFO     flake8.main.application:application.py:265 Reporting errors
INFO     flake8.main.application:application.py:269 Found a total of 4 violations and reported 2
=============================================== short test summary info ===============================================
FAILED tests/test_convention_doc/test_codeblocks.py::test_bad_codeblocks_document_lint_errors[D.1.12-Bad1] - AssertionError: "Bad" codeblock caused unexpected lint error - {'BLK100'}
============================================ 1 failed, 825 passed in 7.41s ============================================
```

CONTRIBUTING.md
```diff
@@ -23,6 +23,16 @@ Please follow these steps to have your contribution considered by the maintainer
...
 ## ⭐️ Common Contributing Workflows ⭐️
 
+## Install Git `pre-commit` Hook Scripts
+
+This project uses [pre-commit](https://pre-commit.com/) to automatically run checks
+when you commit changes to the code. Install the `pre-commit` hook scripts in
+your local copy of the repo by running:
+
+```cmd
+poetry run pre-commit install
```
mshafer-NI, line 33: I'm guessing this is after "poetry install"?

README.md
```diff
@@ -69,3 +69,20 @@ line-length = 100
...
+### Git `pre-commit` Hook Integration
+
+To use [pre-commit](https://pre-commit.com/) to automatically run `ni-python-styleguide` when you commit changes:
+- Add a dev-dependency on `pre-commit`.
+- Add this to your repo's `.pre-commit-config.yaml`:
+
+```yaml
+repos:
+-   repo: https://github.com/ni/python-styleguide/
+    rev: v0.4.1
```
mshafer-NI, line 82: so when we bump, everyone has to bump??

.pre-commit-config.yaml
```diff
@@ -0,0 +1,11 @@
...
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.4.0
+    hooks:
+    -   id: check-toml
+    -   id: check-yaml
+-   repo: https://github.com/psf/black
+    rev: 23.1.0
+    hooks:
+    -   id: black
+        exclude: '^tests/test_cli/acknowledge_existing_errors_test_cases__snapshots/'
```
mshafer-NI, line 11: the `output*.py` files are expected to comply (see pyproject.toml). Are we not able to make this exclude only the input files?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,11 @@
+repos:
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.4.0
+    hooks:
+    -   id: check-toml
+    -   id: check-yaml
+-   repo: https://github.com/psf/black
+    rev: 23.1.0
```
mshafer-NI, line 8: We've chosen to not pin black and instead let the poetry lock pipeline re-run it to update anything needed. Does pre-commit not support just running whatever is in the lock file?

mshafer-NI: Also should we use ni-python-styleguide here which also enforces that we match black's formatting?

mshafer-NI (changes requested):

irwand (approved):
I don't have additional feedback here. I haven't played with pre-commit hook stuff, but the spirit seems good. Thx!

CONTRIBUTING.md
bkeryan: Yes, I'll update this section to say that.

It creates `.git/hooks/pre-commit` with a reference to the currently active Python interpreter:

```
#!/bin/sh
#!/usr/bin/env bash
# File generated by pre-commit: https://pre-commit.com
# ID: 138fd403232d2ddd5efb44317e38bf03

# start templated
INSTALL_PYTHON='C:\Users\bkeryan\AppData\Local\pypoetry\Cache\virtualenvs\ni-python-styleguide-CNd04y3w-py3.9\Scripts\python.exe'
ARGS=(hook-impl --config=.pre-commit-config.yaml --hook-type=pre-commit)
# end templated

HERE="$(cd "$(dirname "$0")" && pwd)"
ARGS+=(--hook-dir "$HERE" -- "$@")

if [ -x "$INSTALL_PYTHON" ]; then
    exec "$INSTALL_PYTHON" -mpre_commit "${ARGS[@]}"
elif command -v pre-commit > /dev/null; then
    exec pre-commit "${ARGS[@]}"
else
    echo '`pre-commit` not found.  Did you forget to activate your virtualenv?' 1>&2
    exit 1
fi
```

Since this is the Python interpreter for my `ni-python-styleguide` virtualenv, it has `pre-commit` installed.

Using `pipx install pre-commit` and `pre-commit install` (with the pipx-installed version) also works.

README.md
mshafer-NI: Then this should be either "main" or "release@latest" (or something) and we start making a "released" branch to reflect what is latest released for use here. (that is, if "use what's in the venv already" isn't an option).

.pre-commit-config.yaml
bkeryan: It supports defining hooks in the local repo: https://pre-commit.com/#repository-local-hooks

It doesn't use the hook definitions in the tool's repo in that case, so you have to specify a few more fields, but perhaps this is worth it to avoid duplicating the tool's version number.

We can use this to run ni-python-styleguide on itself.

.pre-commit-hooks.yaml
```diff
@@ -0,0 +1,7 @@
+-   id: ni-python-styleguide
+    name: ni-python-styleguide
+    description: Enforces NI's internal and external Python conventions.
+    entry: ni-python-styleguide
+    language: python
+    types: [python]
+    args: [lint]
```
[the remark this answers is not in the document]
bkeryan: Good idea. We can certainly define multiple hooks.

bkeryan:
I think it would make sense to run `pre-commit run --all-files` in https://github.com/ni/python-styleguide/blob/main/.github/workflows/PR.yml

bkeryan:
> > I think it would make sense to run `pre-commit run --all-files` in https://github.com/ni/python-styleguide/blob/main/.github/workflows/PR.yml
> 
> Agree with one caveat..., if it makes changes (formatting), how does the developer get those into their local branch? Would they be able to pull? or would they just see that "it got formatted before submission"?

I was thinking it would be the same workflow as when PR.yml runs `ni-python-styleguide lint` directly: the PR build's checks would fail, sending the developer a failure notification email, and they would have to fix the problem manually (possibly by running `pre-commit run --all-files` or `ni-python-styleguide fix`).

`pre-commit` run has more options that seem relevant: https://[pre-commit](https://pre-commit.com/#pre-commit-run).com/#pre-commit-run
- `--show-diff-on-failure`
- `--from-ref` and `--to-ref` would allow restricting to files modified in the PR's branch, but I don't know if that's what we want.

.pre-commit-config.yaml
bkeryan: This exclusion was unnecessary because `pyproject.toml` already excludes those files.

.pre-commit-hooks.yaml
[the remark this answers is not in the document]
bkeryan: Also, `lint` or `fix` should go in `entry` not `args` so that it isn't overridden if the client specifies `args`.

mshafer-NI:
> > > I think it would make sense to run `pre-commit run --all-files` in https://github.com/ni/python-styleguide/blob/main/.github/workflows/PR.yml
> > 
> > 
> > Agree with one caveat..., if it makes changes (formatting), how does the developer get those into their local branch? Would they be able to pull? or would they just see that "it got formatted before submission"?
> 
> I was thinking it would be the same workflow as when PR.yml runs `ni-python-styleguide lint` directly: the PR build's checks would fail, sending the developer a failure notification email, and they would have to fix the problem manually (possibly by running `pre-commit run --all-files` or `ni-python-styleguide fix`).
> 
> `pre-commit` run has more options that seem relevant: https://pre-commit.com/#pre-commit-run
> 
> * `--show-diff-on-failure`
> * `--from-ref` and `--to-ref` would allow restricting to files modified in the PR's branch, but I don't know if that's what we want.

Hmmm, well this impacts whether we should config pre-commit to run "fix" or just "lint" ...
Other than that - sounds good.

.pre-commit-config.yaml
bkeryan: Using local hooks to run Python packages that are already installed in a virtualenv doesn't work very well and the author of `pre-commit` doesn't recommend or support it: 
https://stackoverflow.com/questions/72888074/how-to-configure-pre-commit-config-yaml-to-work-with-poetry

Also, it won't install Python packages located in the current repository:
https://stackoverflow.com/questions/74329631/pre-commit-additional-dependencies-in-a-local-repository

I can easily get `poetry run pre-commit run --all-files` working, but when I run `git commit`, it uses `pre-commit`'s virtualenv, which doesn't have `ni-python-styleguide` installed. 

Using `language: system` and `entry: poetry run ni-python-styleguide lint` doesn't work around this because AFAIK `poetry run` does not activate a virtualenv if one is already active. 

Setting `virtualenvs.in-project: true` and using `entry: .venv\\Scripts\\ni-python-styleguide fix` fails with `Executable `C:devpython-styleguide.venvScriptspython.exe` not found` and I'm not sure why.

mshafer-NI: Well that's fun. Is our best option to make a "release" branch and use that for the git reference?

README.md
bkeryan: I'll update `Publish-Package.yml` to create/update another tag pointing to the new release. Is `latest` okay?

If if we have a `latest` tag, do we need a "released" branch?

Do we care about lightweight vs. annotated tags? It looks like `v0.4.0` is lightweight so I assume we don't care.

mshafer-NI: Like the idea of using a tag.
Wasn't aware there's a difference, I imagine lightweight is fine.

.pre-commit-config.yaml
bkeryan: I think running the previous release of `ni-python-styleguide` will cause problems when the style changes.

The easiest solution for *this* repo is not to run `ni-python-styleguide` from a pre-commit hook. I would like to try using https://github.com/floatingpurr/sync_with_poetry to sync the `black` version.

mshafer-NI: > I think running the previous release of ni-python-styleguide will cause problems when the style changes.

Actually, since ni-python-styleguide doesn't pin `black`, it could also cause a problem with black moves forwards...

I think using that tool to tie the version to poetry.lock is a good optoin.

mbr0wn:
FYI I tried running `pre-commit` with `ni-python-styleguide lint` (so it wouldn't allow me to commit stuff where the style guide checker fails) and I think it's a useful tool. Note I didn't try this branch, I just created a `repo: local` version. I wouldn't mind this getting merged in some shape, form or fashion.

mbr0wn:
Also, if it speeds up things to split the hooks file from the rest, that would be helpful. I'm less interested in developing this tool than in using it.

bkeryan:
I'm closing this PR because I haven't had time to get back to it.

PR #143, opened 2023-10-05
Add support for Python 3.12

### LINKED ISSUE #142: Spurious warnings in f-strings with Python 3.12
With Python 3.12, `ni-python-styleguide` starts emitting E713 for this code:
```python
        if not service:
            raise RuntimeError(
                f"Service class '{service_class}' not found in '{service_config_file}'"
            )
```
```
.\ni_measurementlink_service\measurement\service.py:230:51: E713 test for membership should be 'not in'
```

This is because `ni-python-styleguide` still depends on `pycodestyle` 2.9.1:
- `flake8 = "^5.0"` --> 5.0.4
- `pycodestyle>=2.9.0,<2.10.0` --> 2.9.1

`pycodestyle` 2.11 includes multiple updates for f-string handling. I cannot reproduce this problem with `pycodestyle` main.

bkeryan:

Update version constraints to use `flake8 = "^6.1"` and `pycodestyle = "^2.11"` for Python 3.12 and later. The enhanced f-string support in Python 3.12 changed how the `ast` module parses f-strings, which required updates to `pycodestyle`.

Fix test failures introduced by the update:
- `flake8` no longer tolerates inline comments in the `ignore` field. Comments and error numbers must be on separate lines.
- `pycodestyle` no longer reports "E721 do not compare types, use 'isinstance()'" for comparisons that use the `is` operator.
- `pycodestyle` no longer reports "E741 ambiguous variable name" when using the variable; it only reports this error when assigning the variable.

Update PR workflow to test with Python 3.12.

Fixes #142 

Ran `poetry run pytest -v` with both Python 3.9.13 and 3.12.0.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,7 +16,14 @@ include = ["ni_python_styleguide/config.toml"]
 python = "^3.7"
 
 # Tools we aggregate
-flake8 = "^5.0"
+flake8 = [
+    {version = "^5.0", python = ">=3.7,<3.12"},
+    {version = "^6.1", python = "^3.12"},
+]
+pycodestyle = [
+    {version = "^2.9", python = ">=3.7,<3.12"},
+    {version = "^2.11", python = "^3.12"},
+]
 black = ">=23.1"
 
 # Additional support libraries
--- a/.github/workflows/PR.yml
+++ b/.github/workflows/PR.yml
@@ -46,7 +46,7 @@ jobs:
     strategy:
       matrix:
         os: [macos-latest, windows-latest, ubuntu-latest]
-        python-version: [3.7, 3.8, 3.9, '3.10', '3.11']
+        python-version: [3.7, 3.8, 3.9, '3.10', 3.11, 3.12]
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
--- a/docs/Coding-Conventions.md
+++ b/docs/Coding-Conventions.md
@@ -388,7 +388,7 @@ if cheese is None:
 
 ```python
 # Bad
-if type(num_cheeses) is type(1):
+if type(num_cheeses) == type(1):
     buy(num_cheeses)
 ```
 
--- a/ni_python_styleguide/config.ini
+++ b/ni_python_styleguide/config.ini
@@ -73,18 +73,28 @@ ignore =
     # The same goes for the "descriptive"-mood error, if one was to be added to pydocstyle.
     D401
     # Conflicts with recommending Google style DocStrings
-    D406  # Section name should end with a newline -> forces a newline at end of DocString
+    # D406 - Section name should end with a newline -> forces a newline at end of DocString
+    D406
     # Ignoring things enforced by black
-    D200  # One-line docstring should fit on one line with quotes
-    D206  # Docstring should be indented with spaces, not tabs
-    D207  # Docstring is under-indented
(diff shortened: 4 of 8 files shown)


mshafer-NI (approved):
Thanks for the update!

pyproject.toml
```diff
@@ -16,7 +16,14 @@ include = ["ni_python_styleguide/config.toml"]
...
 # Tools we aggregate
-flake8 = "^5.0"
+flake8 = [
+    {version = "^5.0", python = ">=3.7,<3.12"},
+    {version = "^6.1", python = "^3.12"},
+]
+pycodestyle = [
+    {version = "^2.9", python = ">=3.7,<3.12"},
+    {version = "^2.11", python = "^3.12"},
+]
```
bkeryan, line 26: Note that clients that declare an explicit dependency on `flake8` will not be able to choose a single version that works for Python 3.7 - 3.12. Would it be better for the 1st constraint to be ">=5.0,<7.0"?

@mshafer-NI

mshafer-NI: We don't support client's also requiring `flake8` so I don't see that as much benefit.

mshafer-NI:
> Thank you for contributing! 👋
> 
> Since this changes the Coding Conventions, I'll @-mention the appropriate NI engineers. @alejandro5042 @DavidCurtiss @irwand @jryckman @kcuzner @mshafer-NI @neilvana @nethrasuresh @pakdev @rtzoeller @sbethur @stick152 @tkrebes @Adithyak1998 @innagarc @ShibaniRout

The "conventions change" is just updating an example to something will be flagged with changes to pycodestyle. No effective change.

pyproject.toml
bkeryan: Ok, I tried this.

```
PS D:\dev\measurementlink-python> poetry lock --no-update
Resolving dependencies...

The current project's Python requirement (>=3.8,<4.0) is not compatible with some of the required packages Python requirement:
  - flake8 requires Python >=3.8.1, so it will not be satisfied for Python >=3.8,<3.8.1

Because flake8 (6.1.0) requires Python >=3.8.1
 and no versions of flake8 match >6.1,<7.0, flake8 is forbidden.
So, because ni-measurementlink-service depends on ni-python-styleguide (0.4.1-alpha.0) @ ../python-styleguide which depends on flake8 (^6.1), version solving failed.

  • Check your dependencies Python requirement: The Python requirement can be specified via the `python` or `markers` properties

    For flake8, a possible solution would be to set the `python` property to ">=3.8.1,<4.0"

    https://python-poetry.org/docs/dependency-specification/#python-restricted-dependencies,
    https://python-poetry.org/docs/dependency-specification/#using-environment-markers
```

bkeryan: > We don't support client's also requiring `flake8` so I don't see that as much benefit.

Ok, this is resolved.

mshafer-NI: Yeah, poetry doesn't do great (at least not as of 1.2.2) with bringing compound deps up.

> Changing ni-measurementlink-service to support Python ^3.8.1 instead of ^3.8 allows Poetry to lock flake8 = "*" regardless of the flake8 upper bound specified by ni-python-styleguide.

I would suggest doing `python = "^3.8.1"` along with ni-python-styleguide, and let Poetry figure it out beyond that.

irwand (approved):

PR #148, opened 2023-10-17
Also ignore D102 Missing docstring in public method in tests

irwand:
Some tests are just public method in a test class. We should also ignore this error in tests folder.


mshafer-NI (approved):

mshafer-NI:
Bump the minor in the pyprojec.toml file -> release pipeline is still not doing that.

neilvana (approved):

PR #173, opened 2024-04-04
Ignore F401 for __init__.py files

mbr0wn:
These files routinely import other modules without immediately using them with the intention of making them available to the importing module. Moreover, they are rarely even supposed to be used within the same `__init__.py` file. Marking all of them 'noqa: F410' is redundant.

### DIFF
--- a/docs/Coding-Conventions.md
+++ b/docs/Coding-Conventions.md
@@ -869,7 +869,7 @@ import os  # Assuming os is never used
 
 ```python
 # Good - assuming we are in a __init__.py file
-from .mysubmodule import spam, eggs # OK even if neither are used in this module
+from .mysubmodule import spam, eggs  # OK even if neither are used in this module
 ```
 
 
--- a/docs/Coding-Conventions.md
+++ b/docs/Coding-Conventions.md
@@ -860,10 +860,18 @@ import cheese_shop.brie
 
 > 💻 This rule is enforced by error code F401
 
+`__init__.py` files are an allowed exception because these are used to declare public APIs.
+
 ```python
 # Bad
 import os  # Assuming os is never used
 ```
+
+```python
+# Good - assuming we are in a __init__.py file
+from .mysubmodule import spam, eggs # OK even if neither are used in this module
+```
+
 
 ### [O.1.9] ❌ **DO NOT** Change an imported object's case 💻
 
--- a/ni_python_styleguide/config.ini
+++ b/ni_python_styleguide/config.ini
@@ -112,8 +112,11 @@ ignore =
     # I101 - The names in your from import are in the wrong order. (Enforced by E401)
     I101
 
-# We want to ignore missing docstrings in test methods as they are self documenting
-per-file-ignores= tests/**/test_*.py,tests/test_*.py:D100,D103,D102
+per-file-ignores=
+    # We want to ignore missing docstrings in test methods as they are self documenting
+    tests/**/test_*.py,tests/test_*.py:D100,D103,D102
+    # __init__.py files routinely import other modules without directly using them
+    __init__.py:F401
 
 # Flake8 includes mccabe by default.
 # We have yet to evaluate it, so ignore the errors for now


mshafer-NI:
Have you considered declaring the `__all__` list? this creates a reference while also explicitly declaring the API.
(reference https://docs.python.org/3/tutorial/modules.html#importing-from-a-package)

docs/Coding-Conventions.md
```diff
@@ -858,13 +858,21 @@ import cheese_shop.brie
 
 ### [O.1.8] ❌ **DO NOT** Import definitions that are not used 💻
 
-> 💻 This rule is enforced by error code F401
+> 💻 This rule is enforced by error code F401, except in `__init__.py` files
```
mshafer-NI, line 861: Have you considered declaring the __all__ list? this creates a reference while also explicitly declaring the API.
(reference https://docs.python.org/3/tutorial/modules.html#importing-from-a-package)

mshafer-NI (changes requested):

docs/Coding-Conventions.md
mbr0wn: Hey @mshafer-NI ,

I don't understand the connection to this change request. I'm talking about not having to amend files like this: https://github.com/ni/python-styleguide/blob/main/ni_python_styleguide/_utils/__init__.py

If I understand this correctly, `__all__` is for people doing `from x import *` (which they shouldn't do according to our own [style guide](https://ni.github.io/python-styleguide/#O-1-6)). This doesn't apply to existing `__init__.py` files that already import stuff from submodules.

mbr0wn: @mshafer-NI thanks for sending those explanations, and also the docs on `__all__`, I learned something along the way. 

After processing it for a few days, I still think this here is the right way to go. Also, having scoured peoples thoughts and opinions I really think that `__all__` should be used when `from X import *` does something undesirable, like export functions we want otherwise not exported, or trigger some action that takes long. It should not be used just so we can disable F401 warnings without having to type `noqa: F401`. The downside of using `__all__` is that you now have to maintain lists of exports when doing other changes, which increases maintenance burden and redundancy in `__init__.py` files.

mshafer-NI: Thanks for the response. Two notes:
1. This is a drastic departure from current direction and requires more changes in this PR (this rule becomes an "Avoid" instead of "Do Not", and it needs a lot more examples of when is ok and when is not)
2. We will need buy-in from the rest of the PWG members. I've added this to the agenda for our next meeting on April 17th and will forward you the invite.

ni_python_styleguide/config.ini
```diff
@@ -112,8 +112,11 @@ ignore =
     # I101 - The names in your from import are in the wrong order. (Enforced by E401)
     I101
 
-# We want to ignore missing docstrings in test methods as they are self documenting
-per-file-ignores= tests/**/test_*.py,tests/test_*.py:D100,D103,D102
+per-file-ignores=
+    # We want to ignore missing docstrings in test methods as they are self documenting
+    tests/**/test_*.py,tests/test_*.py:D100,D103,D102
+    # __init__.py files import other modules routinely with directly using them
```
mshafer-NI, line 118:
```suggestion
    # __init__.py files routinely import other modules without directly using them
```

docs/Coding-Conventions.md
mbr0wn: The suggestion I'm making is extremely narrow (just limited to `__init__.py` files), so we should keep the "Do Not" (but I guess we need to update the title). Everything that was previously done to avoid this error (use `noqa` or duplicate symbol names into `__all__`) is still valid and won't throw errors/warnings. I think this won't break the bank 😄

mbr0wn: I updated the language according to our conversation.

irwand (approved):

docs/Coding-Conventions.md
irwand: Thank you for your patience with this.

mshafer-NI (approved):

docs/Coding-Conventions.md
```diff
@@ -860,11 +860,19 @@ import cheese_shop.brie
...
+`__init__.py` files are an allowed exception because these are used to declare public APIs.
+
 ```python
 # Bad
 import os  # Assuming os is never used
 ```
 
+```python
+# Good - assuming we are in a __init__.py file
+from .mysubmodule import spam, eggs # OK even if neither are used in this module
```
mshafer-NI, line 872: Needs to pass `black`

```suggestion
from .mysubmodule import spam, eggs  # OK even if neither are used in this module
```

PR #177, opened 2024-04-28
Update poetry lock and reformat

ni-vsts-admins:
# Update Poetry Lock

Ran:
* `poetry lock`,
* `poetry install`, then
* `ni-python-styleguide fix` to reformat any changes.

\-\-\-
  Created by Github action

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -209,7 +209,7 @@ python-versions = ">=3.7"
 
 [[package]]
 name = "pep8-naming"
-version = "0.13.3"
+version = "0.14.0"
 description = "Check PEP-8 naming conventions, plugin for flake8"
 category = "main"
 optional = false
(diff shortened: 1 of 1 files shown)


irwand (approved):

mshafer-NI:
> @mshafer-NI -- should we take out 3.7 support from this?

hmmmm...., it could cause problems if we did that before all our client packages did.

irwand:
@mshafer-NI : sure but maybe it's a good way to force people to drop 3.7 ? I mean it's dead now 🤷‍♀️

mshafer-NI:
> @mshafer-NI : sure but maybe it's a good way to force people to drop 3.7 ? I mean it's dead now 🤷‍♀️

They'll just use the older package version. That being said, we should drop the validation so the PR can run. I'll post a PR later.

PR #204, opened 2024-11-07
Dev/add bandit linter support

mshafer-NI:
Add [`bandit`](https://bandit.readthedocs.io/en/latest/index.html) as a provided tool.

While `bandit` takes multiple "-c file" parameters, it only uses the latter...
So, to provide a global default (that says it's ok to `assert` in tests, and to skip a lot of non-python folders), AND support for a using project to provide config in their `pyproject.toml`, this is setup to load the vendored-config file, and merge in data from `pyproject.toml` and store the result in a temp file.


irwand (approved):

pyproject.toml
```diff
@@ -26,6 +26,10 @@ pycodestyle = [
     {version = "^2.11", python = "^3.12"},
 ]
 black = ">=23.1"
+bandit = [
+    {version=">=1.5,<1.7", extras=["toml"], python="^3.7,<3.8"},
+    {version=">=1.7.9", extras=["toml"], python="^3.8"},
```
bkeryan, line 31: bandit 1.8.6 requires Python >= 3.9

PR #218, opened 2025-05-02
Drop Python 3.7 support

mshafer-NI:
# Justification
- 3.7 has been EOL since June 27, 2023 (1 year and 11 months)
- We [already stopped testing for it due to limitations of GitHub tooling](https://github.com/ni/python-styleguide/pull/179)
- Both our core tooling (flake8) and most of our dev tooling (notably, pytest) dropped it some time ago.

# Implementation
- bump the 0.x version to denote a breaking change
- declare support for `>=3.8,<4.0` (using the Poetry syntax)

# Testing
PR checks

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,7 +3,7 @@ name = "ni-python-styleguide"
 # The a.0 here denotes a source based version
 # This is removed when released through the Publish-Package.yml GitHub action
 # Official PyPI releases follow Major.Minor.Patch
-version = "0.4.6a0"
+version = "0.5.0a0"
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -13,16 +13,16 @@ include = ["ni_python_styleguide/config.toml"]
 
 
 [tool.poetry.dependencies]
-python = "^3.7"
+python = "^3.8"
```
mshafer-NI, line 16: note, we had already stopped testing 3.7 due to GitHub tooling starting to fail for it.

irwand (approved):

Issue #215, opened 2025-04-14
relative imports in sub-modules are sorted correctly, but flagged incorrectly


mshafer-NI:
While a relative import is in violation of [O.1.5](https://ni.github.io/python-styleguide/#O-1-5), we should flag it for that, rather than for being in the wrong group.

Example:
https://github.com/mshafer-NI/ni-python-styleguide-bug-repro/blob/3fa7c0dada9ee6bb4c4d33e9218905a6c8e4fd97/tests/this/out/test_format.py#L8

Here, we should flag the relative import, and mark the relative import as first party. As is, it is getting flagged as third-party and having an acknowledgement added, but not getting moved to third-party causing confusion on whether or not "fix" did anything.

PR #223, opened 2025-05-07
adopt flake8-tidy-imports to enforce O.1.5

mshafer-NI:
# Reason

We have O.1.5 which says:
> **DO** Use absolute imports

We don't have enforcement of this.

If a non-absolute import is used, our current tooling (`fix` and `format`) has different behavior depending on if it is run above the module (in the folder with `pyproject.toml`) or in a sub-folder within the module. This is (in part) caused by us setting the `app-import-names` automatically if the user does not specify it. (and we don't include neighbor sub-modules).

This PR is a way to address
#215 

Adopting `flake8-tidy-imports` + explicitly setting it to ban "all relative" imports, the case mentioned in that issue is dealt with by flagging that "neighbor" import with I252.

# Pros:

Simple, opinionated rule

# Cons:

(this example assumes a layout similar to):
```
module
|   fizz.py
|   __init__.py
|
\---foo
        __init__.py
```

As [PEP328](https://peps.python.org/pep-0328/#guido-s-decision) defines it, if the import "import foo" is present, it is *considered* to be an absolute import (even though the intention is to resolve to a neighboring sub-module); while, "import .foo" or "from . import foo" are both relative imports. 
**As `flake8-tidy-imports` is currently implemented, this would also ban an "absolute" import that is meant to resolve to a neighbor.**

**Rationale**: 
- Using "import foo" in `fizz.py` is ambiguous for human readers. Do we want `module.foo`? or should `foo` be coming from an install-time dep?
  - If we had further caused confusion by having (e.g.) a `subprocess` sub-module, the code `import subprocess` would resolve to this module's subprocess even if it was intended to resolve to stdlib. By adding this rule in, that would now be required to be `import module.subprocess` or `from module import subprocess`.
- Requiring imports be absolute to the root module (`module`) makes it clear that this is "first-party" and removes ambiguity over whether iSort is doing the right thing or not.


mshafer-NI:
The review-bot was supposed to tag the NI engineers for comment due to this changing the Conventions doc:
@alejandro5042
@DavidCurtiss
@irwand
@jryckman
@kcuzner
@mshafer-NI
@neilvana
@nethrasuresh
@pakdev
@sbethur
@stick152
@tkrebes
@Adithyak1998
@innagarc
@ShibaniRout

irwand (approved):

docs/Coding-Conventions.md
```diff
@@ -802,20 +802,23 @@ from collections import defaultdict
...
 ```
 
-### [O.1.5] ✔️ **DO** Use absolute imports
+### [O.1.5] ✔️ **DO** Use absolute imports 💻
 
 > 🐍 This rule stems from [PEP 8](https://www.python.org/dev/peps/pep-0008)
 
+> 💻 This rule is enforced by error code I252
+
 ℹ️ An exception can be made for `__init__.py` files republishing child module declarations
```
neilvana, line 811: We should either remove this note or add an exception to the linter to allow relative imports from `__init__.py`. I think I'm okay with either.

docs/Coding-Conventions.md
```diff
@@ -869,8 +872,8 @@ import os  # Assuming os is never used
 ```
 
 ```python
-# Good - assuming we are in a __init__.py file
-from .mysubmodule import spam, eggs  # OK even if neither are used in this module
+# Good - assuming we are in a monty/__init__.py file
```
neilvana, line 875: This example isn't really needed if we drop the note, since it is no different from above.

neilvana (changes requested):
I'm in favor of adding the checking. We just need to clarify the exception around `__init__.py`.

kroeschl:
There's a rogue backtick in the PR description:
image

docs/Coding-Conventions.md
mshafer-NI: Our setup is if we have a Bad, we should also have a Good. In this case, it is demonstrating how to do an "absolute" import of the submodule.

PR #238, opened 2025-09-22
github: Upgrade from ubuntu-20.04 to ubuntu-latest

bkeryan:

Upgrade GitHub Actions workflows from `ubuntu-20.04` to `ubuntu-latest`.

`ubuntu-20.04` is no longer supported and is causing workflows to hang, waiting for a runner.

None, I don't have write/admin access.

### DIFF
--- a/.github/workflows/Notify-Convention-Change.yml
+++ b/.github/workflows/Notify-Convention-Change.yml
@@ -12,7 +12,7 @@ jobs:
   org-check:
     name: Check GitHub Organization
     if: github.repository_owner == 'ni'
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


bkeryan:
> Thank you for contributing! 👋

@mshafer-NI @irwand @neilvana That's interesting. Notify-Convention-Change.yml has two jobs, and the job that does the notification already runs on ubuntu-latest. The job that runs on ubuntu-20.04 seems to be intended to restrict the other job to repos in the ni org.

mshafer-NI:
> I wonder if this can be simplified with a job-level `if:`.

Probably This was a workaround to not have comments get posted if I did a draft PR in my fork

neilvana (approved):

PR #260, opened 2026-01-20
disable upgrading flake8-import-order

mshafer-NI:
Renovate keeps trying to upgrade flake8-import-order, which we can't upgrade due to a breaking change that will require work to unblock.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -23,7 +23,7 @@
             ]
         },
         {
-            "matchPackageNames": ["/flake8-import-order/"], // we're not compatible with 0.19.0 yet due to #226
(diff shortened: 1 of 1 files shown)


.github/renovate.json
```diff
@@ -21,6 +21,11 @@
             "extends": [
                 "schedule:weekly"
             ]
+        },
+        {
+            "matchPackageNames": ["/flake8-import-order/"], // we're not compatible with 0.19.0 yet due to #226
```
mshafer-NI: @bkeryan, did I understand correctly that we can use comments in Renovate config files even if they aren't marked as json5 ?

bkeryan (approved):

PR #301, opened 2026-06-12
upgrade poetry lock file to using 2.1.4

mshafer-NI:
We want to use Poetry 2 as standard for our tooling.
Renovate keeps the same version in the lock file, so this change has to be done by hand.

# implementation

`poetry lock`

(without regenerate flag)


irwand (approved):

poetry.lock
```diff
@@ -1,11 +1,12 @@
-# This file is automatically @generated by Poetry 1.8.3 and should not be changed by hand.
+# This file is automatically @generated by Poetry 2.1.4 and should not be changed by hand.
```
bkeryan, line 1: I suggest updating to 2.4.1 and adding a min-release-age to poetry.toml like in https://github.com/ni/hightime/pull/131

mshafer-NI: I want to do that in a separate PR

PR #314, opened 2026-07-10
Ignore linting the .build/ai directory

aluqmanj-ni:
# Justification
Copilot creates ad-hoc scripts in the project's `.build/ai` directory. This is causing linters (ni-python-styleguide, flake8) to flag issues in these scripts when run in the project even though we aren't checking in that code.

# Implementation
Ignore linting the `.build/ai` directory

### DIFF
--- a/ni_python_styleguide/config.ini
+++ b/ni_python_styleguide/config.ini
@@ -117,6 +117,8 @@ per-file-ignores=
     tests/**/test_*.py,tests/test_*.py:D100,D103,D102
     # We don't need to document every Sphinx config
     docs/conf.py:D1
+    # We don't need to document AI scripts
+    .build/ai/**:D1
     # __init__.py files routinely import other modules without directly using them
     __init__.py:F401
 


ni_python_styleguide/config.ini
```diff
@@ -117,6 +117,8 @@ per-file-ignores=
     tests/**/test_*.py,tests/test_*.py:D100,D103,D102
     # We don't need to document every Sphinx config
     docs/conf.py:D1
+    # We don't need to document AI scripts
+    .build/ai/**:D1
```
mshafer-NI, line 121: I'm not sure we want to do this. If it gets checked into source, it should be documented.

bkeryan: Does nps honor .gitignore? is this "AI" stuff typically in .gitignore?

aluqmanj-ni: > is this "AI" stuff typically in .gitignore?

Yes, the `.build` dir is in .gitignore

I've created another PR to update our skills to create AI scripts in `<repo-root>/.build/ai` directory. If that change is approved, this will not be required.

bkeryan: If nps doesn't honor .gitignore, updating it to do so would benefit human developers too.

mshafer-NI: It does not. And while that may be useful, we should include a config option to explicitly include dirs / patterns if we choose to add that.

bkeryan: What do other linters do?

- flake8 does not honor .gitignore: https://github.com/PyCQA/flake8/issues/638
- ruff has a `respect-gitignore` option: https://docs.astral.sh/ruff/settings/#respect-gitignore

aluqmanj-ni: I've created an issue to discuss this further [nps does not honor .gitignore](https://github.com/ni/python-styleguide/issues/319).

Closing this PR since it is no longer needed

PR #327, opened 2026-08-21
Ignore I300 and I301

mshafer-NI:
flake8-import-order recently added I300 and has I301 proposed.

I300 is 

> TYPE_CHECKING block should have one newline above

I301 will probably be 

> TYPE_CHECKING block should have no more than one newline above

&nbsp;

Both of these code conflict with our chosen formatting layout standard of `black`.

So ignore both in favor of `BLK100`

### DIFF
--- a/ni_python_styleguide/config.ini
+++ b/ni_python_styleguide/config.ini
@@ -111,6 +111,9 @@ ignore =
     # flake8-import-order
     # I101 - The names in your from import are in the wrong order. (Enforced by E401)
     I101
+    # I300, I301 - Blank lines around imports and if TYPE_CHECKING. We will use BLK100
+    I300
+    I301
 
 per-file-ignores=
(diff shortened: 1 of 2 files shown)


ni_python_styleguide/config.ini
```diff
@@ -111,6 +111,9 @@ ignore =
     # flake8-import-order
     # I101 - The names in your from import are in the wrong order. (Enforced by E401)
     I101
+    # I300, I301 - Blank lines around imports and if TYPE_CHECKING. We will use BLK100
```
bkeryan, line 114: I300 is "TYPE_CHECKING block should have one newline above"
I301 will probably be "TYPE_CHECKING block should have no more than one newline above"

bkeryan (approved):
Looks good to me, but consider including the warning text in a comment or at least the PR description

